PR #5, opened 2024-11-26
Build cleanup/modernization

nguerrera:
* Use Directory.* props, remove explicit imports and other redundancies from project files.
* Add Directory.Build.rsp so that `msbuild` behaves as `dotnet build` with parallelism and minimal verbosity on by default.
* Simplify build props as much as possible while moving things around. For example, remove unnecessary conditions.
* Use official artifacts layout feature to move all artifacts to bld. I am still getting used to some of the choices it makes here myself, but I think we should give it an honest try before abandoning it because it makes life much better to not have to battle msbuild ourselves every time a new build artifact ends up outside the bld\ folder.
* Put TargetFramework(s) directly in csproj files. Everything in VS breaks without TargetFramework and that's really annoying when working on the build and there's something off. The DRY value here is too low to be worth it, IMHO.
* Enable nullable and implicit usings throughout
* Actually reference Nerbank.GitVersion and Microsoft.SourceLink.GitHub. There was a PackageReference Update that wasn't updating anything.
* Enable .NET Analyzers
* Enable latest analysis level since we're clean on .NET 9 SDK analysis in this repo.
* Update .NET Framework min version from 4.6.2 to 4.7.2. There are various things fixed in tooling by this and I seriously doubt that any partner is actually running on anything lower in 2024. 4.7.2 was released more than six years ago and every Windows release since has that or higher in-box.
* Update .gitignore to latest `dotnet new .gitignore` template + custom exclusion for Directory.Build.rsp

* PR validation improvements:
  * Use release config.
  * Split restore into its own step. Helps when tracking build issues over time to see which step is actually failing easily, and which step is slowing down builds.
  * Don't build again for test step.


src/CSharp/Directory.Build.props
```diff
@@ -1,5 +1,53 @@
...
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <IsPackable>false</IsPackable> <!-- Individual projects must opt-in to packaging -->
+  </PropertyGroup>
+
+  <PropertyGroup Label="SourceLink">
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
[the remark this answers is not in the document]
nguerrera: Thanks, fixed.

src/CSharp/Directory.Build.props
```diff
@@ -1,5 +1,50 @@
...
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/microsoft/cask</RepositoryUrl>
+    <IsPackable>false</IsPackable> <!-- Individual projects must opt-in to packaging -->
+  </PropertyGroup>
+
+  <PropertyGroup Label="SourceLink">
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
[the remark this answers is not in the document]
nguerrera: There are enough tradeoffs to decide this on its own later. I'm definitely not opposed to discussing it.

src/CSharp/Directory.Build.props
```diff
@@ -1,5 +1,50 @@
...
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageProjectUrl>https://github.com/microsoft/cask</PackageProjectUrl>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/microsoft/cask</RepositoryUrl>
+    <IsPackable>false</IsPackable> <!-- Individual projects must opt-in to packaging -->
+  </PropertyGroup>
+
+  <PropertyGroup Label="SourceLink">
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
```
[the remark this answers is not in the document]
nguerrera: https://github.com/dotnet/sourcelink/blob/main/docs/README.md#embeduntrackedsources

It tells the build that if any .cs file is not git controlled, then embed it in the .pdb. This allows stepping through code that's generated in the build. SourceLink pointing to repo alone is not enough for that.

src/CSharp/Directory.WarningsAsErrors.props
```diff
@@ -17,13 +17,5 @@
      -->
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
```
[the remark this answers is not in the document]
nguerrera: There's a comment above that does say what these are (vaguely) and why we suppress (actually we don't suppress, we just don't upgrade them to error like other warnings) them. We're actually clean on them right this minut, but, as expressed in the comment, they're incompatible with warnings-as-errors since they trigger when there's a new vulnerability even if zero code changes have been made. 

I'll make the comment more elaborate and clear. :)

src/CSharp/Test/Directory.Build.props
```diff
@@ -0,0 +1,13 @@
+<Project>
+  <Import Project="..\Directory.Build.props" />
```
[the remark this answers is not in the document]
nguerrera: MSBuildThisDirectory is useless for imports since they are always relative to the current file.

SolutionDir we use elsewhere in project references. I don't like that either because I don't think it's worth having to change the project-relative paths that VS generates automatically on add reference to solution-relative paths. There's a cost if you move a project to a different nesting level in the tree, but that is not an every-day thing.

src/CSharp/Test/Directory.Packages.props
```diff
@@ -0,0 +1,8 @@
+<Project>
```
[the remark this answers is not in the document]
nguerrera: This is a good idea. I think maybe just a separate propertygroup  for those like the build dependencies will work. We can wait until we add one to make the section, I think.

There is also a feature to pin transitive references to versions in this file without explicitly adding a package reference in any .csproj that is interesting for this scenario. I played with it a bit in our other repo but it caused too many issues there. Those issues are arguably things we can/should change, but I had my plate full enough on dependency change risk. :)

src/CSharp/Directory.WarningsAsErrors.props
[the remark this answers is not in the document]
nguerrera: Done. How's that?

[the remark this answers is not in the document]
nguerrera: https://github.com/microsoft/cask/pull/5/commits/b5446b05adfedb3c5a6d548429f4df1883e5df8f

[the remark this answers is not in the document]
nguerrera: Hang on, typos. Lol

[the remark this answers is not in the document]
nguerrera: https://github.com/microsoft/cask/commit/7521e84f0da124a425285c17799556ae47f9b33e

src/CSharp/Test/Directory.Build.props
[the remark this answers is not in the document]
nguerrera: The genuine use case for MSBuildThisFileDirectory is if you are putting it into a property like `<Foo>$(MSBuildThisFileDirectory)..\Bar</Foo>` or something like that.

src/CSharp/Directory.Build.props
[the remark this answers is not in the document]
michaelcfanning: Allows stepping through code and also useful in advanced build determinism/code provenance scenarios (where ad hoc generated builds and packages/pdbs/source are compared).

src/CSharp/Directory.Build.props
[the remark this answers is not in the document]
michaelcfanning: Let's discuss! We have latitude in this project due to its extremely small size. Also valuable to maintain an example that models best practices.

src/CSharp/Directory.Build.props
```diff
@@ -1,5 +1,50 @@
...
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+  </PropertyGroup>
+
+  <ItemGroup Label="Common Packages">
```
michaelcfanning, line 44: >Common Packages [](http://example.com/codeflow?start=20&length=15)

Common build-relevant packages. 

:) NIT. don't block but 'common packages' suggests shared product dependencies to me.

src/CSharp/Directory.Build.rsp
```diff
@@ -0,0 +1,2 @@
+-maxCpuCount
```
michaelcfanning, line 1: >maxCpuCount [](http://example.com/codeflow?start=1&length=11)

was not aware of this feature. nice. and so this is a place for msbuild specific settings that are less desirable to manage in props/targets

src/CSharp/Directory.Build.targets
```diff
@@ -0,0 +1,11 @@
+<Project>
+  <!-- For .NETFramework/.NETStandard allow nullable annotations, but don't enable analysis.-->
+  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+    <Nullable>annotations</Nullable>
+  </PropertyGroup>
+
+  <!-- Generate packages on build for packable projects in release configuration -->
```
michaelcfanning, line 7: >release [](http://example.com/codeflow?start=59&length=7)

ah. i like this.

src/CSharp/Directory.Build.rsp
nguerrera: I use it strictly for command-line switches that can't be expressed in project file. I wouldn't put `-p:Property=Value` in here, that would be evil, IMHO.

src/CSharp/Directory.Solution.props
```diff
@@ -1,3 +1,3 @@
 <Project>
-  <Import Project="targets\build.warnings_as_errors.props" />
+  <Import Project="Directory.WarningsAsErrors.props" />
```
michaelcfanning, line 2: >WarningsAsErrors [](http://example.com/codeflow?start=29&length=16)

is this going to be enabled for our command-line builds? i suppose because we drive build against our SLN, this magic will always happen? 

and so any cmd-line script we build will follow the same build conventions as we have in our actions.

src/CSharp/Cask.Cli/Cask.Cli.csproj
```diff
@@ -1,21 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
-    <Product>Common Annotated Security Keys</Product>
     <TargetFramework>net8.0</TargetFramework>
```
michaelcfanning, line 3: >TargetFramework [](http://example.com/codeflow?start=5&length=15)

so what I infer is that this property is evaluated after first evaluating the directory.build props.

what's the pattern if you happen to need to set something *before* that other props file evaluates?

or do we simply try never to get in that condition? :)

src/CSharp/Cask.Cli/Program.cs
```diff
@@ -2,6 +2,8 @@
 
 using CommandLine;
 
+using System;
```
michaelcfanning, line 5: >System [](http://example.com/codeflow?start=6&length=6)

my care factor is vanishingly small, but why did this come in? our editor settings in other context will flag unused namespace declarations, including `System`. 

do we want to have a standard here? (i.e., express `System` always or allow it to drop). 

eventually, will start investing more in building our our style/correctness checks to help keep code consistent. it is my strong sense that your approach here, driving build with SLN/modern props management will be a big help. we have struggled to keep our in-IDE and command-line style/analyzers enforcements working consistently.

src/CSharp/Directory.Solution.props
nguerrera: Whenever possible, I want consistent behavior between script/action/IDE/dotnet. Putting as much as possible into MSBuild logic gives you that basically free. Making this so for this particular thing took me a while to figure out. Even I didn't know about Directory.Solution.props until a few days ago, lol. But now we have these error codes that are excluded from warning-as-error in one place and it works everywhere :)

src/CSharp/CommonAnnotatedSecurityKeys/Cask.cs
```diff
@@ -67,7 +67,7 @@ public bool IsCask(byte[] keyBytes)
 
         public string GenerateKey(string providerSignature,
                                   string allocatorCode,
-                                  string reserved = null,
+                                  string? reserved = null,
```
michaelcfanning, line 70: >string? [](http://example.com/codeflow?start=34&length=7)

this is surely intentional, since you updated the interface as well, but i don't understand the value. educate me, please.

src/CSharp/Directory.Solution.props
nguerrera: I should note that despite the Directory name, this file isn't automatgic and is imported. I just didn't like having a lonely targets file in its own dir and wanted a name where this one sorts next to the others when I'm making build changes.

It's arguably a bug that I can't paste the contents of Directory.WarningsAsErrors.props into Directory.Build.props and delete Directory.Solution.props.

This bug, in fact: https://github.com/dotnet/msbuild/issues/10801

src/CSharp/CommonAnnotatedSecurityKeys/CommonAnnotatedSecurityKeys.csproj
```diff
@@ -1,38 +1,22 @@
...
-  <PropertyGroup Label="Package">
-    <Title>$(RootNamespace)</Title>
-    <PackageId>$(RootNamespace)</PackageId>
-  </PropertyGroup>
-  
   <ItemGroup>
-    <None Include="../../../docs/CSharp_ReleaseHistory.md" Pack="true" PackagePath="ReleaseHistory.md">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
+    <None Include="../../../docs/CSharp_ReleaseHistory.md" Pack="true" PackagePath="ReleaseHistory.md" CopyToOutputDirectory="true" />
```
michaelcfanning, line 8: what's our best practices for these relative paths? maintain them? locate this file off some other well known location like SLN location? define a var somewhere with BLD_ROOT and/or SRC_ROOT?

src/CSharp/Cask.Cli/Program.cs
nguerrera: Not intentional, implicitusings came and went while I was working, and I must have let something fix this in a bad state. Will revert.

I don't care that much either, but since ImplicitUsings is the default for dotnet new, I'm down to use it everywhere. Alternatively, I would accept not using it anywhere. :)

Prior to this change, we had some projects (probably from dotnet new template) use it and some didn't. That's too confusing, IMHO. I moved it to the props so all projects use it.

src/CSharp/Test/Directory.Build.props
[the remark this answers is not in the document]
michaelcfanning: ok, you partially answered a separate question I had re: use of relative paths. so the open question is whether it makes sense for our path to release history to be anchored relative to some other file.

src/CSharp/CommonAnnotatedSecurityKeys/Cask.cs
nguerrera: I turned on nullable reference types everywhere. This is another thing that only some projects (from dotnet new template) were turning on. It's a great feature, but there is a learning curve.

This is annotating that this parameter may-be-null. Non-null is the default. It's like Spec# if you remember, but reversed ! -> ?. :)

https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references

src/CSharp/Test/Tests.CommonAnnotatedSecurityKeys/TestCaskUtilityApi.cs
```diff
@@ -9,11 +9,8 @@ namespace Tests.CommonAnnotatedSecurityKeys
 {
     internal class TestCaskUtilityApi : CaskUtilityApi
     {
-        public Func<DateTimeOffset> GetCurrentDateTimeUtcFunc;
-
-#pragma warning disable CS0649
-        public Action<byte[], byte[]> ComputeCrc32HashAction;
-#pragma warning restore CS0649
+        public Func<DateTimeOffset>? GetCurrentDateTimeUtcFunc;
+        public Action<byte[], byte[]>? ComputeCrc32HashAction = null; // not currently assigned anywhere else, redundant null assignment to silence compiler warning.
```
michaelcfanning, line 13: >ComputeCrc32HashAction = null [](http://example.com/codeflow?start=39&length=29)

do you understand why we get into this condition for action and not func decl above?

src/CSharp/CommonAnnotatedSecurityKeys/CommonAnnotatedSecurityKeys.csproj
nguerrera: Personally, I prefer to leave them project relative in the project file because that's what VS does when it adds stuff and I don't find much value in fighting it. I don't have strong feelings about this, though.

nguerrera: My other comment elsewhere may read a little stronger than my actual feelings. LOL. :)

src/CSharp/CommonAnnotatedSecurityKeys/CommonAnnotatedSecurityKeys.csproj
```diff
@@ -1,38 +1,22 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
-  <Import Project="$(MsBuildThisFileDirectory)..\Build.props" />
-
   <PropertyGroup>
+    <TargetFrameworks>net8.0;net472</TargetFrameworks>
```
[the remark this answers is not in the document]
nguerrera:
```
$ cmd /c dir /s /b bin obj
D:\Src\cask\bld\bin
D:\Src\cask\bld\obj
```

It does cover that. It also uses some conventions that I don't love.  (More objectively, that am not used to yet. :)). But there's enough difficulty in managing this yourself that I want to dogfood the feature (new as of .NET 8 SDK) for a while to let myself get a more informed opinion. I'm also proud of the .NET SDK team for having the courage (after I left them) to add this feature that was one of our most-requested-but-also-most-bikeshed. So I feel honor-bound to use it. :)

nguerrera:
Hold off on merging, I'll fix the nits.

src/CSharp/Test/Tests.CommonAnnotatedSecurityKeys/TestCaskUtilityApi.cs
nguerrera: Yes, the func was genuinely assigned somewhere, but the action was not.

{CF2C1D18-A3A6-4051-BFF5-D3D7C18C78C3}
{35F89C07-E188-4F4C-981D-418E3DDA0DB5}

The field is public but the class is internal so it's effectively internal. I'm not positive, but it's possible that the C# compiler changed this analysis to whole-assembly at some point. I thought it didn't do this, but it clearly does now, and it's correct here. :)

src/CSharp/Cask.Cli/Cask.Cli.csproj
nguerrera: Yes, there's a pattern, but yes, we also hope to never get in that condition. :)

It's as follows, you can swap the Sdk on the root node for explicit imports of Sdk props and targets that you can put stuff before/after. The Sdk attribute is basically syntactic sugar that expands as below:

``` diff
-<Project Sdk="Microsoft.NET.Sdk">
- ...
-</Project>
+<Project>
+ <!-- Before all common props here, including Directory.Build.props -->
+  <Import Project="Sdk.props" Sdk="Micorosoft.NET.Sdk" />
+  ...
+  <Import Project="Sdk.targets" Sdk="Micorosoft.NET.Sdk" />
+  <!-- After all common targets here, including Directory.Build.targets -->
+</Project>

PR #6, opened 2024-11-26
Use CryptographicOperations.FixedTimeEquals to compare hashes

nguerrera:
This sort of thing was new to me so I chatted with Levi about it to learn more when I saw it. He alerted me to the fact that our implementation was incorrect. 😀

Fixing it, and using this PR as an opportunity to discuss the [polyfill](https://en.wikipedia.org/wiki/Polyfill_(programming)) pattern for .NET Framework that I'm effectively proposing here. I suspect I'm going to need more of this as I go through all the things I've promised to do so it will be good to decide now if we want to do this any differently.

### DIFF
--- a/src/CSharp/CommonAnnotatedSecurityKeys/Cask.cs
+++ b/src/CSharp/CommonAnnotatedSecurityKeys/Cask.cs
@@ -242,26 +242,7 @@ namespace CommonAnnotatedSecurityKeys
         public bool CompareHash(byte[] candidateHash, byte[] derivationInput, byte[] secret, int secretEntropyInBytes)
         {
             byte[] computedHash = GenerateHashedSignatureBytes(derivationInput, secret, secretEntropyInBytes);
-
-            if (computedHash.Length != candidateHash.Length)
-            {
-                return false;
-            }
-
-            bool matched = true;
-            for (int i = 0; i < computedHash.Length; i++)
-            {
-                // We will complete a full comparison of all the data
-                // so that timing differences do not leak information
-                // to an upstream caller (i.e., how much of the hash
-                // matched before we observed a discrepancy).
-                if (computedHash[i] != candidateHash[i])
-                {
-                    matched = false;
-                }
-            }
-
-            return matched;
+            return CryptographicOperations.FixedTimeEquals(computedHash, candidateHash);
         }
     }
 }
--- a/src/CSharp/CommonAnnotatedSecurityKeys/Polyfill/CryptographicOperations.cs
+++ b/src/CSharp/CommonAnnotatedSecurityKeys/Polyfill/CryptographicOperations.cs
@@ -54,7 +54,6 @@ namespace System.Security.Cryptography
 
             return accum == 0;
         }
-
     }
 }
 #endif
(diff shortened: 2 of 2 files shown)


src/CSharp/CommonAnnotatedSecurityKeys/Polyfill/CryptographicOperations.cs
```diff
@@ -0,0 +1,60 @@
...
+            int accum = 0;
+
+            for (int i = 0; i < length; i++)
+            {
+                accum |= left[i] - right[i];
+            }
+
+            return accum == 0;
+        }
+
```
nguerrera, line 57: nit to self: blank line.

src/CSharp/CommonAnnotatedSecurityKeys/Cask.cs
```diff
@@ -242,26 +242,7 @@ internal byte[] GenerateHashedSignatureBytes(byte[] derivationInput, byte[] secr
...
-                // to an upstream caller (i.e., how much of the hash
-                // matched before we observed a discrepancy).
-                if (computedHash[i] != candidateHash[i])
-                {
-                    matched = false;
-                }
-            }
-
-            return matched;
+            return CryptographicOperations.FixedTimeEquals(computedHash, candidateHash);
```
michaelcfanning, line 245: >FixedTimeEquals [](http://example.com/codeflow?start=43&length=15)

Nice! Thanks, @KalleOlaviNiemitalo for shining a light on this concern early in our project work.

src/CSharp/CommonAnnotatedSecurityKeys/Polyfill/CryptographicOperations.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <remarks>
+        ///   This method compares two buffers' contents for equality in a manner which does not
+        ///   leak timing information, making it ideal for use within cryptographic routines.
+        ///   This method will short-circuit and return <c>false</c> only if <paramref name="left"/>
+        ///   and <paramref name="right"/> have different lengths.
+        ///
+        ///   Fixed-time behavior is guaranteed in all other cases, including if <paramref name="left"/>
+        ///   and <paramref name="right"/> reference the same address.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
```
michaelcfanning, line 34: >MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization [](http://example.com/codeflow?start=20&length=63)

sure! hadn't occurred to me but totally makes sense.

src/CSharp/CommonAnnotatedSecurityKeys/Polyfill/CryptographicOperations.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                return false;
+            }
+
+            int length = left.Length;
+            int accum = 0;
+
+            for (int i = 0; i < length; i++)
+            {
+                accum |= left[i] - right[i];
```
michaelcfanning, line 52: >accum |= left[i] - right[i]; [](http://example.com/codeflow?start=16&length=28)

nice! also wouldn't have occurred to me. must get a micro-optimization here that's useful.

the more tuned this expression is, the less the impact from having to process all N bytes.

security attackers make it tough to get around. please see devo's `we're through being cool` for more on this.

PR #7, opened 2024-11-26
Use file-scoped namespaces

nguerrera:
Save some indentation. Better to touch every line sooner than later. 😊

Reviewing individual commits may be helpful.

1. `dotnet format` with current .editorconfig
2. update .editorconfig to use file-scoped namespaces, `dotnet format` again.
3. delete commented out file because it confused me that count of .cs files touched didn't match expectation.


nguerrera:
Hide whitespace option also helpful for review.

src/CSharp/Cask.Benchmarks/CompareHashedSignatureBenchmarks.cs
```diff
@@ -5,63 +5,62 @@
 
 using System.Security.Cryptography;
 
-namespace CommonAnnotatedSecurityKeys.Benchmarks
+namespace CommonAnnotatedSecurityKeys.Benchmarks;
```
michaelcfanning, line 8: >namespace [](http://example.com/codeflow?start=0&length=9)

`If it were done when 'tis done, then 'twere well it were done quickly`

michaelcfanning (approved):
:shipit:

PR #9, opened 2024-12-02
Re-organize folder structure

nguerrera:
1. Remove CSharp folder

It's really annoying to go yet another level deeper all the time IMHO. We may decide to put C++ in the same sln anyway, and we can probably put other language project as siblings of C# projects whether they are in a sln or not.

2. Rename Test folder to Tests

It's more consistent with project/namespace names.

3. Move Benchmarks to Tests

This is a natural generalization of "Tests" meaning IMHO to include perf testing. It further allows one directory for non-product build props.

4. Move *.md from source to new root devdocs folder

These have language-agnostic info and shouldn't be in the C# project.

5. Rename projects for consistency and brevity:

* CommonAnnotatedSecurityKeys -> Cask
* Tests.CommonAnnotatedSecurityKeys -> Cask.Tests

We still use CommonAnnotatedSecurityKeys as the namespace.


nguerrera:
I have a big PR coming that builds on this structure. I want to ensure the history gets the renames and the diff in the big PR isn't too hard to review with renames getting lost.

PR #11, opened 2024-12-02
Second draft

nguerrera:
## Singleton/Interface -> Static methods
* This mirrors the modern .NET Crypto API upon which these are built.
* Eliminates the thread-static lazy pattern that was incorrect.
* Eliminates the need for callers to think about lifetimes or threads at all.
* Move `ICask` cross-language API to test project. Tests will implement this interface with some adapting to their target language's native API.

## Add a strongly typed `CaskKey` struct.
- This is just a string wrapper. No performance overhead vs. string after creation.
- Guaranteed to be valid Cask unless default-initialized with null string.

## Spanify
* We no longer allocate any memory on the heap except to produce the string to back a new CaskKey.
* We support UTF-16, UTF-8, and decoded byte inputs as spans now. You can still pass strings too.
* We put limits on the sizes of key components so that we can `stackalloc` unconditionally and make strong assertions like above on how much we allocate.

## Analyzers and formatting
* Start fresh from `dotnet new editorconfig`. 
* Bump "suggestions" to "warning" to make dotnet format fix them.
* Turn on all .NET analyzers, opt-out of a few rules individually.
* Enforce `dotnet format --verify-no-changes` in PR Validation.
* It's very possible that we will decide to change `.editorconfig` for different style choices or to opt-out of more rules.

This PR does not come with strong opinions on these settings, only the strong opinion that whatever setting we choose should be enforced. I've opted for an aggressive starting point because it's always easier to dial this back than to dial it forward.

## Establish patterns, principles, and guidelines for .NET Framework support

See lengthy comment in `Polyfill.cs`.

## Also
* Add more argument validation.
* Rename "reserved" to "provider data". 
* Fix various issues found by analyzers.
* Use .NET 9 SDK in PR validation.
* Build and test debug and release in PR validation. (For Debug.Assert coverage, mainly.)


src/.editorconfig
```diff
@@ -160,7 +160,9 @@ indent_size = 4
 tab_width = 4
 
 # New line preferences
-end_of_line = crlf
+# VS and dotnet format conspire to require us to pick one so we let 2/3 supported OSes win.
+# .gitattributes is configured to make this work on Windows too
+end_of_line = lf
```
nguerrera, line 165: I tried deleting this line. VS puts it back. I tried using `unset`, ditto.

nguerrera: I've been down this road before with JS, didn't expect it with C#.

If you get weird issues due to the switch, this PR has instructions:

https://github.com/Azure/azure-sdk-for-js/pull/10360

nguerrera: An alternative here is to disable the line-ending rule, but actually after getting over my frustration, I think this is better since it makes the repo resilient to different user git configurations. Also, windows tools are fine with LF these days, even notepad. :)

src/Cask/Helpers.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            if (!IsValidForBase64Url(c))
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static bool IsValidForBase64Url(char c)
```
nguerrera, line 72: We can do better than this perf-wise. Base64Url.IsValid from the BCL package we now use is more heavily optimized. However, it allows whitespace and padding. And as I write this, it occurs to me that we probably want a stricter validation in IsCask to stop whitespace and padding!

Feels like something worth leaving to its own change.

src/Directory.Build.props
```diff
@@ -1,18 +1,25 @@
 <Project>
   <PropertyGroup Label="Build">
     <LangVersion>latest</LangVersion>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <AnalysisLevel>latest</AnalysisLevel>
+    <AnalysisMode>all</AnalysisMode>
     <Deterministic>true</Deterministic>
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
```
nguerrera, line 8: This combined with deterministic makes it so that if you don't change surface area, the build won't recompile everyone who depends on you. Super nice.

.github/workflows/validate.yml
```diff
@@ -18,16 +18,27 @@ jobs:
...
+      run: dotnet test src -c Debug --no-build
+
+    - name: Build (Release)
       run: dotnet build src -c Release --no-restore
 
-    - name: Test
+    - name: Test (Release)
       run: dotnet test src -c Release --no-build
+
+
```
nguerrera, line 44: nit to self: extra lines

devdocs/Questions.md
```diff
@@ -1,15 +1,65 @@
...
+   - Cranked up the analyzers to eleven and looked where they led me. This was a great exercise even for me because I've been away from .NET for four years I learned a ton. Learning new platform features this way made me very nostalgic for FxCop. :)
+   - We will surely decide to turn off some analyzers as we go and that's fine.It's as easy as Ctrl-dot -> Configure Severity -> Silent.
+
+1. I got rid of the CSharp\ folder. It was driving me nuts to have everything one level deeper than ususal. 
+   - As we mentioned before we may put C++ in the same sln. 
+   - Even if not, I think we can make non-C# projects siblings of C# projects.
+
+1. Dialed back the number of test cases for timestamp. This code path is too simple to spend that many cycles testing on every developer iteration.
+
+1. Surely lots I'm forgetting, but I'll add to here as we go over live.
```
nguerrera, line 59: So many typos here, but this file isn't meant to last beyond the code review.

Putting a comment so I remember to delete it before merging.

src/Directory.Build.targets
```diff
@@ -1,7 +1,9 @@
 <Project>
-  <!-- For .NETFramework/.NETStandard allow nullable annotations, but don't enable analysis.-->
-  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <!-- For .NETFramework/.NETStandard allow nullable annotations, but don't enable analysis.-->
```
nguerrera, line 3: delete .netstandard from comment

src/Directory.Packages.props
```diff
@@ -1,17 +1,19 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
```
nguerrera, line 4: This one is for @rwoll from his prior feedback. Tested it out and it works great. You can lift a transitive dep by adding a version in this file and leaving the .csproj files alone. This still implicitly creates new top-level dependencies, but the technical debt is at least constrained to this one file. I made a section for the debt below.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -20,129 +23,223 @@ protected CaskTestsBase(ICask cask)
...
 
-        IsCaskValidate(Cask, key);
+        IsCaskValidate(key);
+    }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    [InlineData("A")]   // Too short
+    [InlineData("ABC")] // Too lng
```
nguerrera, line 98:
```suggestion
    [InlineData("ABC")] // Too long
```

devdocs/Questions.md
nguerrera: Sorry, this is an existing file and I will not delete it, but I will remove my ramblings from it.

src/Cask/CaskKey.cs
```diff
@@ -0,0 +1,123 @@
...
+        if (!TryEncode(bytes, out CaskKey key))
+        {
+            ThrowFormat();
+        }
+        return key;
+    }
+
+    public override string ToString()
+    {
+        return _key ?? "";
```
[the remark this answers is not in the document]
nguerrera: Great point! Since we can't (musn't) throw in ToString(), how about this:

1. A different API to get the string that does throw when uninitialized.
2. A DebuggerDisplayAttribute to show you the string in the debugger.

src/Cask/CaskKey.cs
```diff
@@ -0,0 +1,123 @@
...
+
+namespace CommonAnnotatedSecurityKeys;
+
+/// <summary>
+/// Represents a Cask key or hash.
+/// </summary>
+public readonly partial record struct CaskKey
+{
+    // PERF: Do not add more fields. The layout here is intentionally identical
+    // to that of a string reference.This means that this type provides type
```
nguerrera, line 17:
```suggestion
    // to that of a string reference. This means that this type provides type
```

src/Cask/Polyfill.cs
```diff
@@ -0,0 +1,275 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+/*
+ * https://en.wikipedia.org/wiki/Polyfill_%28programming%29 
+ *
+ * "In software development, a polyfill is code that implements a new standard
+ * feature of a deployment environment within an old version of that environment
+ * that does not natively support the feature.
```
nguerrera, line 9:
```suggestion
 * that does not natively support the feature."
```

src/Cask/Polyfill.cs
```diff
@@ -0,0 +1,275 @@
...
+ *    for good?" Write that code, and then come here and make it work somehow on
+ *    .NET Framework. We want the pain of .NET Framework contained to this file
+ *    that we'll hopefully delete one day when .NET Framework support can be
+ *    dropped.
+ *
+ *  - Check if there are official out-of-band .NET Microsoft.Bcl.* or System.*
+ *    that support .NET Framework before attempting to write it yourself.
+ *
+ *  - Use the Polyfill namespace when types exist on .NET Framework, but are
+ *    missing functionality.Use extension methods to add instance methods, and
```
nguerrera, line 37:
```suggestion
 *    missing functionality. Use extension methods to add instance methods, and
```

src/Directory.Build.targets
nguerrera, line 3:
```suggestion
    <!-- For .NETFramework, allow nullable annotations, but don't enable analysis.-->
```

src/Tests/Cask.Tests/CSharpCaskTests.cs
```diff
@@ -1,11 +1,70 @@
...
+
+    private sealed class Implementation : ICask
     {
+        public bool CompareHash(byte[] candidateHash,
+                                byte[] derivationInput,
+                                byte[] secret,
+                                int secretEntropyInBytes = 32)
+        {
+            //return CSharpCask.CompareHash(candidateHash, derivationInput, secret, secretEntropyInBytes);
+            throw new NotImplementedException();
```
nguerrera, line 24: Left-over from mid-development. There are no unit tests for these functions yet, clearly. :)

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -20,129 +23,223 @@ protected CaskTestsBase(ICask cask)
...
+    [MemberData(nameof(TooShortOrLongForAKey))]
+    public void CaskSecrets_IsCask_InvalidKey_Basic(string? key)
+    {
+        // We need helpers to make it easier to create keys that are "nearly
+        // valid" as in valid except in one dimension like length. We have an
+        // example of this IsKeyValidate where we put back a valid checksum
+        // after modifiying a key, but it needs to be easier to reuse in more
+        // specific tests. It's hard because the IsValid check has a lot of
+        // redunancy (not a bad thing!). For example, if you change the length
+        // it will can fail alignment, not just checksum. This test and similar
```
nguerrera, line 50:
```suggestion
        // it can fail alignment, not just checksum. This test and similar
```

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
+        destination = destination[1..];
+
+        // Copy provider data
+        ReadOnlySpan<byte> source = secretBytes[secretEntropyInBytes..];
+        source[..providerDataLengthInBytes].CopyTo(destination);
+        source = source[providerDataLengthInBytes..];
+        destination = destination[providerDataLengthInBytes..];
+
+        // TODO: Pseudo-code says allocator code from key with new timestamp?
+        // But how can we expect hashes to compare equal if each hash gets a new timestamp?...
```
nguerrera, line 246: I wrote this comment while still confused that the timestamp only changes when the month rolls over but it's still a problem.

src/Cask.Cli/CloudInstance.cs
```diff
@@ -1,6 +1,9 @@
-﻿namespace Cask.Cli;
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-public enum CloudInstance
+namespace CommonAnnotatedSecurityKeys.Cli;
+
+internal enum CloudInstance
```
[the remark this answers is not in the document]
nguerrera: Because it's in the Cli project. If/when we want to make public API out of this, it should also be moved to the Cask project.

As for why I bothered to change this now, there is an analzyer that says "why are you putting public API in your application project" and I actually agree with it.

devdocs/Questions.md
```diff
@@ -1,15 +1,65 @@
...
-1. Code is broken for dates in 2050 forward (need a test injection mechanism to unit test)
-1. Create a language neutral interface declaration
-1. CompareHash should use heavily optimized MemoryExtensions.SequenceEqual
-1. Tests should have their own root folder so they can have their own D.*.props
-1. Can we eliminate InternalsVisibleTo?
 1. Add hard-coded keys for testing.
-1. Fluent assertions vs. xunit assertions (we choose xunit for now)
 1. Stress, concurrency, performance, fuzzing, RNG behavior testing.
+1. Code coverage reporting in CI
+1. Unit tests for generate/compare hash
```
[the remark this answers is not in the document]
nguerrera: 100%. There wass already a TODO for this a few lines above :)

src/Cask/Cask.csproj
```diff
@@ -2,21 +2,30 @@
...
-    <PackageReference Include="System.Collections.Immutable" />
+  <ItemGroup>
+    <InternalsVisibleTo Include="Cask.Tests" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Folder Include="Properties\" />
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
+    <Compile Remove="Polyfill.cs" />
+    <None Include="Polyfill.cs" />
```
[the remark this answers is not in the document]
nguerrera: Yeah, I had started with a polyfill *directory*, but when I realized how many I was going to need I couldn't bring myself to have that many files just for .NET Framework stubs so I put them all in one file. :)

src/Cask/Polyfill.cs
```diff
@@ -0,0 +1,275 @@
...
+ *    only resort to shadowing types to add static methods.
+ *
+ *  - Use the real BCL namespace to backport types that don't exist at all on
+ *    .NET Framework.
+ *
+ * - Only resort to #if if you have exhausted all other options.
```
nguerrera, line 43: Missing guideline: keep everything in this file `internal`

src/Tests/Cask.Benchmarks/CompareHashedSignatureBenchmarks.cs
```diff
@@ -33,33 +33,18 @@ public void UseCompareHash()
...
-        string key = Cask.Instance.GenerateKey(TestProviderSignature, "99");
-        byte[] keyBytes = Convert.FromBase64String(key.FromUrlSafe());
-        string hash = Cask.Instance.GenerateHash(TestProviderSignatureBytes, keyBytes, 32);
-        var hmac = new HMACSHA256(keyBytes);
-        byte[] hashBytes = hmac.ComputeHash(TestProviderSignatureBytes);
+        CaskKey key = Cask.GenerateKey(TestProviderSignature, "99");
+        byte[] keyBytes = Base64Url.DecodeFromChars(key.ToString().AsSpan());
+        byte[] hashBytes = new byte[HMACSHA256.HashSizeInBytes];
+        HMACSHA256.HashData(keyBytes, s_testProviderSignatureBytes, hashBytes);
+        Span<byte> computedHashBytes = stackalloc byte[32];
```
nguerrera, line 40: nit to self: be consistent here with too lines above.
```suggestion
        Span<byte> computedHashBytes = stackalloc byte[HMACSHA256.HashSizeInBytes];
```

devdocs/Questions.md
```diff
@@ -1,15 +1,65 @@
...
+   - Reserving a byte or two for future use by us migt be a good idea anyway so two bytes of padding may be OK? See next question for case in point.
+
+1. It would be useful to be able to determine the boundary between secret and provider data. Can we write the secret entropy length somewhere?
+  - Having to pass the secret entropy length to GenerateHash along seems like a footgun.Ideally, you only need to specify this to create a key and never to use it, IMHO.
+
+# NGuerrera's Pre-Code Review Notes
+1. I'm sorry, this huge and bordering on a rewrite. I got a little carried away over the holiday. I should have broken this up into smaller pieces. Just like we did for MikeFan's the first draft, we should go over this live together.
+
+1. On checksum question above: CRC32 is a good choice, we're using it for it's intended purpose. MD5 will just trigger false alarms about insecure algorithms. 
+   -BUT, can we make room for all four bytes? There isn't any prior art on truncating CRC-32 like SHAs.
```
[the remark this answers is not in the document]
michaelcfanning: The issue here is that any lack of alignment across a 4 byte boundary causes issues. First, arbitrary division of bits in the encoded key means you can't always author minimal regexes (or any regex) for a specific pattern. That isn't an issue with a checksum itself but it does impact the encoded strings leading up to them. If we allow a full checksum to spill over into the end of a key, then we either need to permit the `=` character (which has to be encoded in a URL further complicating regex patterns) *or* we put a new requirement on callers to trim these characters.

This design, like many, is a balance of concerns, which means there is something to criticize for nearly every feature. If we expand the size of certain components, that is contrary to some users who want the smallest key size possible. If we make it easy to encode at the byte level, we compromise readers who want to treat the string textually, etc. I am not saying there are no conversations to be had, exactly the contrary. But we should ultimately document the full set of trade-offs and all the various scenarios and use cases (which include efficiency & simplicity of minting keys, transmitting/storing them, validating them, examining them for literal security purposes and the small matter of detecting and preventing their unwanted exposure.

src/Directory.Build.props
michaelcfanning: Nice! We should push this into other projects, it's a nice optimization for local build/testing.

src/Tests/Cask.Tests/CSharpCaskTests.cs
```diff
@@ -1,11 +1,70 @@
...
+        }
+
+        Mock ICask.MockFillRandom(FillRandomAction fillRandom)
+        {
+            return CSharpCask.MockFillRandom(fillRandom);
+        }
+
+        Mock ICask.MockUtcNow(UtcNowFunc getUtcNow)
+        {
+            return CSharpCask.MockUtcNow(getUtcNow: getUtcNow);
```
michaelcfanning, line 67: >getUtcNow: getUtcNow [](http://example.com/codeflow?start=41&length=20)

NIT: I tend not to repeat a named parameter when names match up. The devil, of course, is when you provide a single parameter of multiple (out of order) that requires you to provide both. :)

nguerrera: Strong agree! VS must have done this when I did a refactor :)

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
+        Debug.Assert(maxLength < MaxKeyLengthInBytes);
+        Span<byte> keyBytes = stackalloc byte[maxLength];
 
-        return
-            crc32Bytes[0] == keyBytes[keyBytes.Length - 3] &&
-            crc32Bytes[1] == keyBytes[keyBytes.Length - 2] &&
-            crc32Bytes[2] == keyBytes[keyBytes.Length - 1];
-    }
+        OperationStatus status = Base64Url.DecodeFromChars(
+            key,
```
michaelcfanning, line 49: >key [](http://example.com/codeflow?start=12&length=3)

Another style NIT we need to close on. We have preferred an old C++ style convention where the first parameter it left on the same line as the function call, with all other params dropped down and left aligned to the first. We should stick with that style or argue about it here and change the standard. Personally I prefer the former because it (to me) reads more naturally by co-locating the parameters close to the function/method name.

nguerrera: I prefer the latter because:

1. I can't figure out how to make `dotnet format` do it for me.
2. I regularly find myself spending time manually moving things around after a function name changes length.

I will not die on this hill.

src/Cask/Helpers.cs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public static bool Is4CharAligned(int charLength)
+    {
+        return charLength % 4 == 0;
+    }
+
+    public static int GetKeyLengthInBytes(int secretEntropyInBytes, int providerDataLengthInBytes)
+    {
+        Debug.Assert(Is3ByteAligned(secretEntropyInBytes), "secretEntropyInBytes should have been rounded up to 3-byte alignment already.");
```
michaelcfanning, line 54: >Is3ByteAligned [](http://example.com/codeflow?start=21&length=14)

I see why you handled it this way and I agree with the it. The caller here really needs to have established these lengths otherwise they can't process outputs properly.

src/Cask/Helpers.cs
```diff
@@ -0,0 +1,111 @@
...
+
+global using static CommonAnnotatedSecurityKeys.Helpers;
+
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace CommonAnnotatedSecurityKeys;
+
+internal static class Helpers
```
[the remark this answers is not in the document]
nguerrera: These shouldn't be extension methods, regardless of namespace or pollution. 

It's not idiomatic to have someValue.DoSomeMath() in .NET.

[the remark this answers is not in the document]
nguerrera: With `using static` as I have done, I don't see anything wrong with `DoSomeMath(someIntegerValue)`.

Extension methods are very often overused IMHO.

devdocs/Questions.md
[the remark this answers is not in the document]
nguerrera: I want two more bytes right now:

1 for full CRC32 from partial
1 for secret entropy length

This leaves a single unused byte to get back to alignment. I think that would be acceptable and it may even be wise to have a reserved byte.

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
-                                           allocatorCode,
-                                           reservedBytes,
-                                           secretEntropyInBytes);
+        if (status != OperationStatus.Done)
+        {
+            return false;
+        }
 
-        return Convert.ToBase64String(keyBytes).ToUrlSafe();
+        keyBytes = keyBytes[..bytesWritten];
```
michaelcfanning, line 63: is this necessary? you already compute the max decoded length, and you verified the decoding operation completed. aren't we done after `DecodeFromChars`? i.e., why would `bytesWritten` differ from `maxLength`, am I missing something?

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
+    /// <summary>
+    /// Validates that the provided UTF8-encoded byte sequence represents a valid Cask key.
+    /// </summary>
+    public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUTF8)
     {
-        // Ensure that the randomBytesCount is a multiple of 3. We keep all data
-        // aligned along a 3-byte boundary to ensure consistent base64 encoding
-        // in the key for fixed components.
-        secretEntropyInBytes = secretEntropyInBytes.RoundUpToMultipleOf(3);
+        // NOTE: Since all valid Cask keys are ASCII-safe, we use the same UTF-16 char limits as byte limits..
```
michaelcfanning, line 72: >UTF-16 [](http://example.com/codeflow?start=75&length=6)

did you mean UTF8 here? also note the repeated periods at the end of this comment.

nguerrera: No I mean UTF-16 because `char` is a UTF-16 code unit. This comment is meant to explain why we have `byte`s in hand but are consulting `char` limits. I'll try to word it better.

src/Cask/Cask.cs
nguerrera: I suppose it could throw here now.

nguerrera: Actually, I can also use our own prediction about the correct length instead of max-length.

I will make this better.

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
+        if (keyUTF8.Length < MinKeyLengthInChars || keyUTF8.Length > MaxKeyLengthInChars || !Is4CharAligned(keyUTF8.Length))
+        {
+            return false;
+        }
 
-        byte[] allocatorAndTimestampBytes = GenerateAllocatorAndTimestampBytes(allocatorCode);
+        ReadOnlySpan<byte> signature = keyUTF8[^16..^12];
```
michaelcfanning, line 78: >[^16..^12] [](http://example.com/codeflow?start=46&length=10)

So, note this CASK magic. It remains to be seen whether we retain this advantage but currently you can examine a well-known location from the end of the key to make a very fast determination that you've got a CASK thing.

This only works because don't allow any optional data at the end of the key. This is what causes our uncertainty around the literal entropy of the key (because there could be optional data following the literal randomized component). 

If we did *not* have a fixed location for this signature and looked for it, things would be more complex, we'd skip the first encoded 24 chars (because this is the minimal supported 128 bit randomized component) and keep checking 4 byte segments for `JQQJ`. An issue here is that this approach would require us to advise key creators to never allow `JQQJ` to appear in their optional component, an unwanted bit of friction to worry about.

We could simply require a 256 randomized component and then the `JQQJ` identifier is again in an entirely fixed location,  but this comes at the cost of expanding the encoded key size to support a level of entropy that isn't strictly required today to be unbreakable.

We could only support specific key sizes (in terms of entropy), e.g., you can allocate a 128 bit, 256, 384 or 512 bit key. Now we could look at the key length to assess the entropy and find the fixed `JQQJ` component (but only if the key allocate hasn't allocated enough optional data to force us to erroneously conclude the entropy is greater than it is. This limit would be 21 optional encoded characters, 24 additional bits could fool us into thinking someone had allocated 128 bits more entropy than they had).

More detail here than might be expected on this topic but all the issues/considerations above are still in play.

My specific comment here is that the magic numbers here probably warrant capturing as constants. :) And possibly the standard signature itself `JQQJ`. Oof. This comment is a long haul to a small note.

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
 
-        if (testChar == default)
+        if (status != OperationStatus.Done)
         {
-            CaskUtilityApi.Rng.GetBytes(keyBytes);
+            return false;
         }
-        else
+
+        keyBytes = keyBytes[..bytesWritten];
```
michaelcfanning, line 103: >bytesWritten [](http://example.com/codeflow?start=30&length=12)

Same note, do we need this?

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
-            Array.Copy(Convert.FromBase64String(randomComponent), 0, keyBytes, 0, secretEntropyInBytes);
+            return false;
         }
 
-        int reservedOffset = secretEntropyInBytes;
-        int caskSignatureOffset = reservedOffset + reservedLength;
-        int allocatorAndTimestampOffset = caskSignatureOffset + 3;
-        int providerSignatureOffset = allocatorAndTimestampOffset + allocatorAndTimestampBytes.Length;
-        int partialHashOffset = providerSignatureOffset + providerSignature.Length;
+        // Check signature: [0x25, 0x04, 0x09] is JQQJ decoded from base64 to bytes
```
michaelcfanning, line 121: >bytes [](http://example.com/codeflow?start=78&length=5)

declare a const somewhere? We have `Limits`, do we need `Signatures` or some other notion? We have char ranges for the timestamp values in particular that also might be useful to capture elsewhere. I mention this because considering the range of fixed data may help you think about naming and factoring. Another consideration, of course, is that key implementers will design their own optional data and look to our example for a model of how to process it.

src/Cask/Cask.cs
```diff
@@ -1,247 +1,412 @@
...
+        if (0x25 != signature[0] || 0x04 != signature[1] || 0x09 != signature[2])
+        {
+            return false;
+        }
 
-        // Copy optional reserved bytes, if provided.
-        Array.Copy(reserved ?? new byte[] { }, 0, keyBytes, reservedOffset, reserved?.Length ?? 0);
+        // Validate 3-byte partial CRC32 checksum
+        ReadOnlySpan<byte> keyBytesWithoutCrc = keyBytes[..^3];
+        ReadOnlySpan<byte> crc = keyBytes[^3..];
```
michaelcfanning, line 130: >keyBytes [](http://example.com/codeflow?start=33&length=8)

Just another observation, more magic here as we have a fixed location + length for the checksum. the fixed location of the standard signature and the checksum make the detection easy/performant while achieving the accuracy levels we're looking for.

src/Cask/Cask.cs
nguerrera: Fixed.

src/Cask/Cask.cs
michaelcfanning, line 80: >ReadOnlySpan [](http://example.com/codeflow?start=8&length=12)

ok, so just observing, we perform the encoded signature check early as a perf micro-optimization so we don't take the costs of decoding into bytes to perform the check. fine. i wonder what the actual performance savings are.

you know, as I look at this, I realize that only the last 12 bytes are relevant to the validation with the critical note that you need to process the key in its entirely to produce the checksum.

it raises the (potentially bad) idea of only applying the checksum to the first 9 bytes of this 12 byte range (i.e., the checksum would exclude the randomized and optional components leading up to the end of the key).

if we did this, we'd only need to decode those final 12 bytes. we'd look for the signature in the first 3 bytes, we'd compute the CRC32 of the first 9 bytes, and look for the first 3 bytes of the CRC32 in the final 9 bytes.

the result is that any key issued for the same provider signature, allocated in the same month/year would have the same checksum. and this sequence could be appended to arbitrary other data and we'd see the key as valid. so there's a trade-off.

but what are we resisting here? we'd still achieve the same goals with the approach above of eliminating false positives in most practical circumstances. what we'd also do is lower the costs of spoofing keys.

nguerrera: Another option:

``` diff
< Final 3 bytes = [CRC_0] [CRC_1] [CRC_2] [CRC_3]
> Final 6 bytes = [RESERVED_BY_CASK_SPEC] [ENTROPY_LENGTH] [CRC_0] [CRC_1] [CRC_2] [CRC_3]
```

Pros:
- We gain the full CRC32
- We retain fixed offsets from the end
- We retain variable-length entropy and variable length provider data
- API can take `CaskKey` alone instead of `(CaskKey key, int secretEntropyInBytes)`

Cons:
- 3 more bytes/4 more chars per key

nguerrera: Oh, and I will move the magic numbers to constants. Good call.

Just also contributing to the long haul discussion. :)

devdocs/Questions.md
```diff
@@ -1,15 +1,30 @@
 # Questions
 1. MD5 vs. CRC32 (or something else)?
-1. Should we have a dedicated return type?
-1. Timing attacks in computed hash?
+   - CRC32 is a good choice, we're using it for it's intended purpose.
```
nguerrera, line 3:
```suggestion
   - CRC32 is a good choice, we're using it for its intended purpose.
```

devdocs/Questions.md
```diff
@@ -1,15 +1,30 @@
...
+1. Should provider data be allowed to be unaligned. Should we pad it instead of throwing?
+
+1. Should we combine Cask and CaskKey, putting the statics as helpers on the struct?
+  - Best name for combined? Or better names (plural) if they should remain separate?
+  - Related: can we avoid repeating Cask like Cask.IsCask -> Cask.IsValid.
+
+1. Should we use a new timestamp in GenerateHash and mask it out in CompareHash?
+   - The first draft created a new timestamp, but didn't mask so it would fail if the month rolled over between GenerateHash and CompareHash.
+   - The second draft copies the timestamp from the secret when generating a hash.
+   - It feels wrong for something called "GenerateHash" to not be deterministic so I think copying is better.
```
nguerrera, line 18:
```suggestion
   - It feels wrong for something called "GenerateHash" to not be deterministic.
```

src/Cask/CaskKey.cs
```diff
@@ -0,0 +1,123 @@
...
+        return key;
+    }
+
+    public override string ToString()
+    {
+        return _key ?? "";
+    }
+
+    // language=regex
+    private const string RegexPattern = """(^|[^A-Za-z0-9+/-_])([A-Za-z0-9-_]{4}){6,}JQQJ[A-Za-z0-9-_]{12}($|[^A-Za-z0-9+/-_])""";
```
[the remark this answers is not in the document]
nguerrera: Since this is moved from a different location, and this change is already too large, I will leave this to another change to improve on it.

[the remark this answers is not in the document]
nguerrera: I appended to TODO on this.

src/Tests/Directory.Packages.props
```diff
@@ -1,8 +1,8 @@
 <Project>
   <Import Project="..\Directory.Packages.props" />
-  <ItemGroup Label="Test-Only Dependencies">
-    <PackageVersion Include="BenchmarkDotNet" Version="0.13.12" />
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
+  <ItemGroup>
+    <PackageVersion Include="BenchmarkDotNet" Version="0.13.2" />
```
nguerrera, line 4:
```suggestion
    <PackageVersion Include="BenchmarkDotNet" Version="0.13.12" />
```

src/Tests/Cask.Tests/CSharpCaskTests.cs
nguerrera, line 67:
```suggestion
            return CSharpCask.MockUtcNow(getUtcNow);
```

src/Cask/Cask.cs
nguerrera: I made a TODO for constants and I will fast-follow-up with that.

src/Cask/Cask.cs
nguerrera: I made a TODO for constants and I will fast-follow-up with that.

src/Cask/Cask.cs
nguerrera: I made a TODO for constants and I will fast-follow-up with that.

devdocs/Questions.md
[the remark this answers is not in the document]
nguerrera: Resolving because this is an open question, recorded as one, and not changed by this PR.

src/Cask.Cli/GenerateCommand.cs
```diff
@@ -1,45 +1,41 @@
...
-        byte[] reserved = Convert.FromBase64String($"{cloudText}{region}{tenant}{providerReserved}");
+        string providerData = $"{cloudText}{region}{tenant}{providerReserved}";
 
-        byte[] providerSignature = Convert.FromBase64String(options.FixedSignature.ToUrlSafe());
+        string providerSignature = options.FixedSignature;
 
         for (int i = 0; i < options.Count; i++)
         {
+            CaskKey key = Cask.GenerateKey(providerSignature,
+                                           "99",
```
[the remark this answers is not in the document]
nguerrera: Only because this PR is huge, I appended to TODO on this. I will fast follow-up on it.

src/Cask/CaskKey.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace CommonAnnotatedSecurityKeys;
+
+/// <summary>
+/// Represents a Cask key or hash.
+/// </summary>
+public readonly partial record struct CaskKey
```
michaelcfanning, line 14: nice! moving forward we will likely want to have helpers on this struct that help return things like the time stamp (perhaps converted to an actual date time), the optional data etc. 

I like this but what can we do to create a pattern where someone creates their FooCaskKey that can utilize as much goodness as possible.

src/Cask/Helpers.cs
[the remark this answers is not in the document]
nguerrera: We talked offline and we're going to leave this.

src/Cask/CaskKey.cs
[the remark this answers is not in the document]
nguerrera: Added to TODO. We talked offline about challenges here. It will be better to discuss on its own PR.

nguerrera:
I've taken this out of draft, but marked it with no-merge label to prevent it from merging. I added the action to check the no-merge label. The only reason I'm marking this no-merge is that I'd like to review the commit message and make sure it's not a concat of my mess of commits.

src/Cask/Helpers.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    public const int FixedKeyComponentSizeInBytes =
+        3 +  // CASK signature
+        3 +  // Allocator code and timestamp
+        3 +  // Provider signature
+        3;   // Checksum
+
+    public const string Base64UrlChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
+
+    public static int RoundUpTo3ByteAlignment(int bytes)
```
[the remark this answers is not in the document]
nguerrera: Yeah, I'm not sure either. That reads a little awkwardly to me too. I'll leave it and if we come up with a better scheme, I'll be happy to change it.

src/Cask/Cask.cs
```diff
@@ -1,247 +1,415 @@
...
+        // NOTE: Since all valid Cask keys are ASCII-safe, we use the we can
+        //       safely apply `char` limits as `byte` limits.
+        if (keyUTF8.Length < MinKeyLengthInChars || keyUTF8.Length > MaxKeyLengthInChars || !Is4CharAligned(keyUTF8.Length))
+        {
+            return false;
+        }
 
-        byte[] allocatorAndTimestampBytes = GenerateAllocatorAndTimestampBytes(allocatorCode);
+        ReadOnlySpan<byte> signature = keyUTF8[^16..^12];
+        if ('J' != signature[0] || 'Q' != signature[1] || 'Q' != signature[2] || 'J' != signature[3])
```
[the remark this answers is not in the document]
nguerrera: There's performance and also the risk of leading people to try to use UTF-8 API on bytes that are actually not text but base64-decoded bytes.

I will look at reducing the repetition, but it will need some thought, so I added it to TODO. I can probably come up with something better.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -20,129 +23,232 @@ protected CaskTestsBase(ICask cask)
...
+        // after modifiying a key, but it needs to be easier to reuse in more
+        // specific tests. It's hard because the IsValid check has a lot of
+        // redunancy (not a bad thing!). For example, if you change the length
+        // it can fail alignment, not just checksum. This test and similar
+        // trivial tests below were stepped through to check code coverage of
+        // current implementation, but they are susceptible to starting to pass
+        // for the wrong reason if/when implementation changes.
+
+        bool valid = Cask.IsCask(key!);
+        Assert.False(valid, $"'IsCask' unexpectedly succeeded with an invalid key: {key}");
```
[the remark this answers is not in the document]
nguerrera: We can't, there's a lengthy comment above about improving this.

.github/workflows/validate.yml
```diff
@@ -18,16 +18,25 @@ jobs:
...
       run: dotnet --info
 
     - name: Restore
       run: dotnet restore src
 
-    - name: Build
+    - name: Check Formatting
+      run: dotnet format --verify-no-changes src
+
+    - name: Build (Debug)
```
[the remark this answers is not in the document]
nguerrera: Added to TODO.

src/Cask/Cask.cs
```diff
@@ -1,247 +1,415 @@
...
-        // This check ensures that 3 bytes of fixed signature are present
-        // where they belong. We next hash the key and ensure the first
-        // three bytes of the hash are present where they belong. This
-        // leads to a chance of collision of 1 in 2 ^ 48, or
-        // 1 in 281,474,976,710,656, or ~1 million times less likely than
-        // winning the Powerball lottery.
-        if (CaskUtilityApi.CaskSignatureBytes[0] != keyBytes[keyBytes.Length - 12] ||
-            CaskUtilityApi.CaskSignatureBytes[1] != keyBytes[keyBytes.Length - 11] ||
-            CaskUtilityApi.CaskSignatureBytes[2] != keyBytes[keyBytes.Length - 10])
+        ReadOnlySpan<char> signature = key[^16..^12];
```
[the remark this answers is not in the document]
nguerrera: Added to TODO.

src/Tests/Cask.Benchmarks/IsCaskBenchmarks.cs
```diff
@@ -1,32 +1,35 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Buffers.Text;
```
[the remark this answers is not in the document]
nguerrera: Added to TODO.

src/Tests/Cask.Tests/CaskSecretsTests.cs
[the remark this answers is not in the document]
nguerrera: And sorry, I missed "Perhaps this is what your comment above is alluding to?"

There are two sides to this. The comment is more about how it's not that easy to make a key that's invalid for only one reason.

The other side is that the API doesn't report why it's invalid, it just returns a bool, which maybe we should change. Needs design.

src/Cask/Cask.cs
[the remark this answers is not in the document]
nguerrera: Added to TODO.

src/Tests/Cask.Tests/CaskSecretsTests.cs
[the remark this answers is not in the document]
nguerrera: Added to TODO.

PR #14, opened 2024-12-11
Set up VS Code developer experience

nguerrera:
## Basics
* Configure default solution
* Configure default build task: Ctrl+Shift+B -> Build
* Configure debug launch tasks: F5 -> Debug
* Configure basic formatting options to match .editorconfig
* Recommend EditorConfig and C# DevKit extensions
* Exclude a few things from the explorer tree

## Format-on-save

Enable format-on-save for C# code in VS Code.

Since we enforce formatting in PR validation, this won't cause untouched code to reformat. And for new code, it helps you pass PR validation without an extra formatting step.

I wish I could set this up for big VS too, but it doesn't seem to be possible yet: https://github.com/dotnet/roslyn/issues/76256

## Also
* Move .editorconfig to root to apply to files outside src
* Specify UTF-8 without BOM for all files by default
* Specify two space indentation for everything but C# files
* Use CRLF and/or UTF-8 with BOM when big VS insists on saving that way
* Add additional files to solution items
* Remove BOMs to match settings above
* Fix newlines in SUPPORT.md
* Fix a few stray files that were still missing a final newline
* Fix typo in Directory.Build.props comment
* Fix minor formatting inconsistency between workflow yaml files


src/Directory.Build.props
```diff
@@ -50,7 +50,7 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
-  <!-- Share .NET Framework polyfilss with other projects -->
+  <!-- Share .NET Framework polyfills with other projects -->
```
michaelcfanning, line 53: >polyfills [](http://example.com/codeflow?start=28&length=9)

my god, how did i miss this?

PR #16, opened 2024-12-12
Initial skeleton code for C++ Implementation

nguerrera:
* Add libcask native dll project to solution
* Add stub implementation aligned to test ICask interface
* Add P/Invoke interop in tests
* Use MSBuild and Xunit magic so that we
  1. Have a good experience in VS on Windows for C++
  2. Don't regress the C# dev experience x-plat for C#

The focus here is on getting the build off the ground so folks can work in parallel. The API shape will surely change when an actual non-stub C++ implementation is worked on.

This currently works only on Windows x64 and only with big VS, but everything is set up there to the point where you can drop a breakpoint in .cpp and hit it from the Test Explorer.

More work will be needed to match this experience for Linux, Mac, VS Code on Windows, and non-x64 architectures.

There is also no PR validation coverage yet because that will require infrastructure to build with big VS and/or solving x-plat.

### Also (not strictly related)
* Revert back to CRLF on Windows

There are too many issues with VS putting snippets with CRLF into files and making them have mixed line endings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,5 @@
-# Use unix line endings always, even on Windows
-*                            text=auto eol=lf
-
-# Exceptions to above for files that VS saves with CRLF always
-*.sln                        eol=crlf
-*.lutconfig                  eol=crlf
+# Normalize line endings to LF in repo, checkout CRLF on Windows
+*                            text=auto
 
 # Allow comments in JSON in GitHub rendering
 *.json                       linguist-language=JSON-with-Comments
--- a/src/Cask.sln
+++ b/src/Cask.sln
@@ -16,6 +16,8 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Build", "Build", "{0C3A2105-9369-461A-92AB-3D39CA120B83}"
 	ProjectSection(SolutionItems) = preProject
 		..\.editorconfig = ..\.editorconfig
+		..\.gitattributes = ..\.gitattributes
+		..\.gitignore = ..\.gitignore
 		Cask.lutconfig = Cask.lutconfig
 		Directory.Build.props = Directory.Build.props
 		Directory.Build.rsp = Directory.Build.rsp
@@ -51,6 +53,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Workflows", "Workflows", "{
 		..\.github\workflows\validate.yml = ..\.github\workflows\validate.yml
 	EndProjectSection
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libcask", "libcask\libcask.vcxproj", "{14013CD3-B963-4851-AA9A-7C7A2F110A52}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -73,6 +77,8 @@ Global
 		{FB74046B-2FF6-4316-85B1-39A28D945A18}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{FB74046B-2FF6-4316-85B1-39A28D945A18}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{FB74046B-2FF6-4316-85B1-39A28D945A18}.Release|Any CPU.Build.0 = Release|Any CPU
+		{14013CD3-B963-4851-AA9A-7C7A2F110A52}.Debug|Any CPU.ActiveCfg = Debug|x64
+		{14013CD3-B963-4851-AA9A-7C7A2F110A52}.Release|Any CPU.ActiveCfg = Release|x64
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/src/Tests/Cask.Tests/Cask.Tests.csproj
+++ b/src/Tests/Cask.Tests/Cask.Tests.csproj
@@ -1,9 +1,32 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net8.0;net472</TargetFrameworks>
+    <OSIsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</OSIsWindows>
+    <OSIsX64 Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">true</OSIsX64>
+    <MSBuildIsNETFramework Condition="'$(MSBuildRuntimeType)' == 'full'">true</MSBuildIsNETFramework>
+    <BuildCpp Condition="'$(OSIsWindows)' == 'true' and '$(MSBuildIsNETFramework)' == 'true' and '$(OSIsX64)' == 'true'">true</BuildCpp>
+    <!-- TODO: Our custom C++ build logic is breaking fast-up-to-date for this project. -->
+    <DisableFastUpToDateCheck>true</DisableFastUpToDateCheck>
+    <EnableUnmanagedDebugging>true</EnableUnmanagedDebugging>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\Cask\Cask.csproj" />
+    <None Update="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(BuildCpp)' == 'true'">
+    <CppProjectReference Include="..\..\libcask\libcask.vcxproj" Properties="Platform=x64" />
+    <Content Include="$(ArtifactsPath)\bin\libcask\$(Configuration.ToLowerInvariant())_x64\libcask.dll" CopyToOutputDirectory="PreserveNewest" Visible="false" />
+    <AssemblyMetadata Include="BuiltWithCppSupport" Value="true" />
+  </ItemGroup>
+
+  <Target Name="BuildCppProjectReference" BeforeTargets="DispatchToInnerBuilds">
+    <MSBuild Projects="@(CppProjectReference)" />
+  </Target>
+
+  <Target Name="CleanCppProjectReference" BeforeTargets="Clean" Condition="'$(TargetFramework)' == ''">
+    <MSBuild Projects="@(CppProjectReference)" Targets="Clean" />
+  </Target>
+
 </Project>
--- a/src/Tests/Cask.Tests/Cask.Tests.csproj
+++ b/src/Tests/Cask.Tests/Cask.Tests.csproj
@@ -21,7 +21,7 @@
     <AssemblyMetadata Include="BuiltWithCppSupport" Value="true" />
   </ItemGroup>
 
-  <Target Name="BuildCppProjectReference" BeforeTargets="DispatchToInnerBuilds">
+  <Target Name="BuildCppProjectReference" BeforeTargets="DispatchToInnerBuilds" Condition="'$(InnerTargets)' == 'Build'">
     <MSBuild Projects="@(CppProjectReference)" />
   </Target>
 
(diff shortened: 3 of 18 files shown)


suvamM:
@nguerrera Taking a look at this PR.

nguerrera:
Thanks, I have it in draft while I fight the build. I think it's going to work now.

src/Tests/Cask.Tests/Cask.Tests.csproj
```diff
@@ -1,9 +1,32 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net8.0;net472</TargetFrameworks>
+    <OSIsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</OSIsWindows>
+    <OSIsX64 Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">true</OSIsX64>
+    <MSBuildIsNETFramework Condition="'$(MSBuildRuntimeType)' == 'full'">true</MSBuildIsNETFramework>
+    <BuildCpp Condition="'$(OSIsWindows)' == 'true' and '$(MSBuildIsNETFramework)' == 'true' and '$(OSIsX64)' == 'true'">true</BuildCpp>
```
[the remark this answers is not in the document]
nguerrera: (Disclaimer: I'm rusty on the details here and it's complicated.)

> or is everything a string type in MSBuild?

No, properties can have other runtime types in msbuild internals. Quotes don't necessarily turn things into strings either. 'true' is still a boolean, but if you compare 'true' to 'potato' then that will become a string comparison. It's like MSBuild watched https://www.destroyallsoftware.com/talks/wat and said, "Hold my beer!".

What you wrote will only work if you ensure that every boolean property is definitively assigned a boolean value. It is very common to have properties that are either 'true' or '' in practice. Which is the case here because I assign true conditionally but I never assign false. This is idiomatic in this weird language, as is comparing to 'true'.

src/libcask/.gitignore
```diff
@@ -0,0 +1 @@
+libcask.sln
```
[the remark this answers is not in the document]
nguerrera: OMG, this was not intended to be checked in. There is no libcask.sln. I went down some dark roads before I got to this final state.

src/Cask.sln
```diff
@@ -51,6 +53,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Workflows", "Workflows", "{
 		..\.github\workflows\validate.yml = ..\.github\workflows\validate.yml
 	EndProjectSection
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libcask", "libcask\libcask.vcxproj", "{14013CD3-B963-4851-AA9A-7C7A2F110A52}"
```
michaelcfanning, line 56: >libcask [](http://example.com/codeflow?start=64&length=7)

my care factor is not high, but i'm not sure I've seen 'lib' as a prefix for building something like this, is it a convention in your experience?

src/libcask/cask.cpp
```diff
@@ -0,0 +1,73 @@
...
+    //      challenge for C API and I'm open to other approaches.
+
+    const char* key = "not_a_real_key";
+    int32_t requiredSizeInBytes = int32_t(strlen(key) + 1); // +1 for null terminator
+
+    if (output == nullptr)
+    {
+        // no buffer: return the minimum size of the buffer needed to succeed.
+        // caller can then allocate a buffer of that size and call again.
+        return requiredSizeInBytes;
```
michaelcfanning, line 41: >requiredSizeInBytes [](http://example.com/codeflow?start=15&length=19)

we are just getting started here, so no need to think of this feedback as blocking, but I'm not sure that this convention of telling the caller the size of the buffer makes sense, as the size is easily computable from the arguments provided to us.

I suppose that we do have things like our fixed signature and checksum, however, which might not be obvious. So, ok, we could provide this data. In which case your implementation isn't correct because we aren't accounting for the size of the input data, e.g., the provider data.

Anyway, I know you're not intending to provide an actual implementation, you're simply proposing a pattern, which is to allow people to:

1) Provide a null out argument, in which case the API providers the computed final size of the encoded key,
2) Caller then allocates space for the encoded key and calls again.

src/Cask.sln
nguerrera: It is the common convention on Unix-y platforms: libthing.so. Eventually it will definitely be the best name for the .so when we have one. 

The (Core) CLR actually has magic so that you can PInvoke to "foo" and it will try libfoo.so. This lets you call it `thing.dll` on Windows and `libthing.so` elsewhere and your portable C# PInvoke can do `[DllImport("Foo")]` and work with both.

But we already have cask.dll (right now, we can change that, of course) for managed implementation. I could call it cask_native.dll but that's ugly and libcask_native.so or cask_native.so would be more offensive to my eyes than libcask.dll is.

I am not going to insist on dll/so names, but that's how I got to this point. We can certainly talk about better names.

nguerrera: I suppose the most common convention is:

Big.Long.Namespace.For.Thing.dll - Managed Code
thing.dll - Native Windows code
libthing.so - Native Mac/Linux code

src/libcask/cask.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+     return false; 
+}
+
+CASK_API int32_t Cask_GenerateKey(const char* allocatorCode,
+                                  const char* providerSignature,
+                                  const char* providerData,
+                                  int32_t secretEntropyInBytes,
+                                  char* output,
+                                  int32_t outputSizeInBytes)
```
michaelcfanning, line 28: >outputSizeInBytes [](http://example.com/codeflow?start=42&length=17)

oh boy. we have a very old school C-style WIN32 API style api here.

we might prefer in our implementation to use std::string (and convert to c_str() as required on return) in the implementation itself just to have a cleaner/more correct implementation. or we could just knock it out old school i guess.

src/libcask/cask_dependencies.cpp
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// WIP: This is our reference implementation for dependencies. We can pull in
+//      external libraries of our choice here and only here. We can use C++
+//      here.
+
+#include "cask_dependencies.h"
+
+std::string Cask::Base64UrlEncode(const std::span<uint8_t>& bytes)
```
michaelcfanning, line 10: >Base64UrlEncode [](http://example.com/codeflow?start=18&length=15)

we need a fill bytes routine and also something for retrieving a timestamp, yes?

src/Cask.sln
nguerrera: Our namespace without a dot is throwing me though: CommonAnnotatedSecurityKeys. Also, I'm feeling a little rebellious on the long dll and folder names. It's hard to come back to them after four years away from managed code.

src/Tests/Cask.Tests/CppCaskTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+[assembly: DefaultDllImportSearchPaths(DllImportSearchPath.SafeDirectories | DllImportSearchPath.UseDllDirectoryForDependencies)]
+
+namespace CommonAnnotatedSecurityKeys.Tests;
+
+// WIP: These tests are disabled because the C++ implementation is stubbed out.
+//      To enable them, flip the return value of IsSupportedTestClass in
+//      TestFilter.cs.
+
+public class CppCaskTests : CaskTestsBase
```
michaelcfanning, line 34: >CppCaskTests [](http://example.com/codeflow?start=13&length=12)

i guess we can benchmark against this test wrapper. but you could imagine a reference implementation of C# wrapper against a C++ low level implementation could make sense in the product itself. 

that is, if the C++ performance waxes the C#. :) the gauntlet is thrown down! prove that there is no rationale for productizing C# interop to a non-C# impl!
heh

src/libcask/cask.cpp
nguerrera: Yep, I literally said, I'm open to other ideas in the comment just above this. This is always my least favorite part of defining C API.

My size is "correct" here because I always return "not_a_real_key" in this placeholder. This is the size required to provide that. When this actually generates a key, it will need to do the computation as a function of the inputs. If it did this, then the current double-pinvoke dance would still work to get a key of the right size. 😊

Among the other ways to do this annoying chore:

1. Separate GetKeyLength/GetHashLength API. These would be fine by me.
1. Document the math and make people do it themselves
1. Take a callback function pointer and passes you back (via reverse p/invoke if you are managed). 
I hate all of them, but we can choose our poison later.

src/libcask/cask.cpp
nguerrera: I think Michael means we use std:string internally and copy from c_str() to the user buffer as a final step .

I think it would make a lot of sense for this layer to be a marshalling layer in and out of a nicer C++ layer. I started to pontificate about that in my "WIP: here's what you're gonna wanna do." comments but then erased it as it wasn't well written and winning my battle with the build took the energy I had left. :)

src/Tests/Cask.Tests/CppCaskTests.cs
nguerrera: I don't think there's any chance that the interop will win.

nguerrera: This test wrapper isn't a good benchmark, it is not even trying to be fast. The cross-language interface still used arrays and such.

It is conceivable that a non-trivial amount of effort could produce an interesting race between portable C# and C# -> C++ interop.

But then I still doubt that the latter would win by enough of a margin to make it worth having to deal with issues of operating systems and architectures forever when talking to our C# customers. I think it's rational to pay some perf to get rid of that.

I'm personally comfortable not bothering to "prove" it. If you look at our benchmarks now, in private chat, but PRs coming, this thing is going to be damn fast already. I don't think it makes sense to trade portability away for faster-than-really-fast.

src/libcask/cask_dependencies.cpp
nguerrera: Free as in there's nothing technically stopping it from working. Depending how carried away the C++ gets, there could be perf or other costs.

.gitattributes
```diff
@@ -1,9 +1,5 @@
-# Use unix line endings always, even on Windows
-*                            text=auto eol=lf
-
-# Exceptions to above for files that VS saves with CRLF always
-*.sln                        eol=crlf
-*.lutconfig                  eol=crlf
+# Normalize line endings to LF in repo, checkout CRLF on Windows
+*                            text=auto
```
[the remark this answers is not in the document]
nguerrera: I've said that if I had a time machine, I'd prevent the invention of more than one line-ending scheme. I'd also prevent a second slash direction and the concept of capital letters. :)

src/Cask.sln
michaelcfanning: Is there a convention? Yes. asked and answered, thanks. :)

src/libcask/cask.cpp
michaelcfanning: I think the pattern you've pointed us to is fine.

src/libcask/libcask.vcxproj
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
nguerrera: Yes. This is on my radar! I will make x-plat great, I promise.

nguerrera:
@KalleOlaviNiemitalo Docs will come very soon. That said, this project is just getting off the ground and you may continue to see things hit the repo out of natural order.

src/libcask/cask.cpp
nguerrera: Yes, good point. We probably want to reserve a place for error codes across the board and consistently. And we probably want this irrespective of whether or not they are needed by the initial implementation.

nguerrera:
> please get a sign off from someone with production C++ experience (which does not include me)!

@michaelcfanning OK to merge?

michaelcfanning (approved):
:shipit:

PR #18, opened 2024-12-19
Performance and benchmark work

nguerrera:
* Improve benchmarks
  * Don't do more than one iteration within a benchmark, let BenchmarkDotNet handle that.
  * Support BenchmarkDotNet CLI options and picking benchmarks to run interactively.
  * Add benchmarks for key and hash generation.
  * Use consistent naming conventions for benchmarks.
  * Improve benchmark debugging experience.
  * For each benchmark, where applicable, add a 'floor' counterpart to measure what doing the same scenario with non-identifiable keys would cost. We can't do better than this and it establishes how much it costs to use Cask in exchange for the additional value it provides.

* Improve .NET Framework Polyfills
  * Reduce allocations.
  * Add tests borrowed from BCL and a small threading stress test.
  * Exclude polyfill attributes that are not instantiated via reflection from code coverage.
  * Don't copy polyfill code into test/benchmark assemblies, use it via InternalsVisibleToAttribute. Improves code coverage experience.

* Misc
  * Don't allocate a new string when creating a CaskKey from an existing string.
  * Rename CaskKey.(Try)Parse to CaskKey.(Try)Create. Parsing isn't really what we do.
  * Limit length of derivation input.
  * Add fast convenience API for providing derivation input as a string.
  * Remove open question about timestamp in hash, since this is settled. Update pseudo-code, and add test.
  * Add gitignore entry for Ultra profiling files.
  * Fix some exception messages.
  * Add a bit more test coverage.
  * Set a maximum stackalloc constant of 256 bytes and assert that it is never exceeded.
  * Move test-specific warning suppression to test-specific .editorconfig for better interaction with `dotnet format`.


nguerrera:
Benchmark results on my machine.

## .NET 8.0

| Method             | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------------- |-----------:|----------:|----------:|-------:|----------:|
| CompareHash_Cask   | 451.609 ns | 3.9292 ns | 3.6753 ns |      - |         - |
| CompareHash_Floor  | 294.875 ns | 2.7652 ns | 2.5866 ns |      - |         - |
| GenerateHash_Cask  | 300.978 ns | 1.3669 ns | 1.2786 ns | 0.0091 |     176 B |
| GenerateHash_Floor | 225.569 ns | 1.6476 ns | 1.2864 ns | 0.0057 |     112 B |
| GenerateKey_Cask   | 143.634 ns | 2.8027 ns | 2.7527 ns | 0.0093 |     176 B |
| GenerateKey_Floor  |  63.337 ns | 1.2881 ns | 1.6290 ns | 0.0058 |     112 B |
| IsCaskString       |  23.718 ns | 0.1039 ns | 0.0921 ns |      - |         - |
| IsCaskBytes        |   7.501 ns | 0.0291 ns | 0.0272 ns |      - |         - |

Notice how nothing but the bytes to hold the generated key or hash is allocated on the heap. :)

## .NET Framework with naive polyfills (before optimizations in this PR)

| Method             | Mean       | Error    | StdDev   | Gen0   | Allocated |
|------------------- |-----------:|---------:|---------:|-------:|----------:|
| CompareHash_Cask   | 2,601.8 ns | 15.03 ns | 14.06 ns | 0.1450 |     931 B |
| CompareHash_Floor  | 1,809.6 ns | 16.60 ns | 13.87 ns | 0.0801 |     514 B |
| GenerateHash_Cask  | 2,627.9 ns | 50.53 ns | 69.17 ns | 0.2022 |    1292 B |
| GenerateHash_Floor | 1,752.4 ns | 12.52 ns | 11.71 ns | 0.1163 |     738 B |
| GenerateKey_Cask   | 1,080.1 ns |  5.02 ns |  4.45 ns | 0.1583 |    1003 B |
| GenerateKey_Floor  |   561.4 ns |  2.69 ns |  2.39 ns | 0.1297 |     818 B |
| IsCaskString       |   181.4 ns |  0.38 ns |  0.33 ns |      - |         - |
| IsCaskBytes        |   128.9 ns |  0.90 ns |  0.85 ns |      - |         - |

## .NET Framework with optimized polyfills (with optimizations in this PR)

| Method             | Mean       | Error    | StdDev   | Gen0   | Allocated |
|------------------- |-----------:|---------:|---------:|-------:|----------:|
| CompareHash_Cask   | 2,542.0 ns | 36.60 ns | 34.23 ns | 0.0534 |     353 B |
| CompareHash_Floor  | 1,802.4 ns |  8.94 ns |  7.93 ns | 0.0458 |     289 B |
| GenerateHash_Cask  | 2,320.9 ns | 10.17 ns |  9.01 ns | 0.1106 |     714 B |
| GenerateHash_Floor | 1,741.5 ns |  7.54 ns |  7.06 ns | 0.0801 |     514 B |
| GenerateKey_Cask   |   707.3 ns |  8.41 ns |  7.87 ns | 0.0725 |     457 B |
| GenerateKey_Floor  |   184.6 ns |  0.76 ns |  0.71 ns | 0.0446 |     281 B |
| IsCaskString       |   175.9 ns |  3.50 ns |  3.28 ns |      - |         - |
| IsCaskBytes        |   128.1 ns |  0.64 ns |  0.60 ns |      - |         - |

Obviously, .NET Framework remains much slower than .NET 8.0. The remaining allocations are not in our control. In some cases, I think GC is just not that significant relative to the other increased costs of things we're doing on .NET Framework (like crypto just being significantly slower) and so reducing bytes allocated isn't making things that much better. In other cases, we have decent gains here. 

My starting point was let's not optimize .NET Framework at all. We'll say it works, but if you care about perf, use modern .NET. But now I think it's not terrible to do the best we can as long as we can continue to keep it contained to polyfills and not spread `#if` all over the place.

## Legend
```
  Mean      : Arithmetic mean of all measurements
  Error     : Half of 99.9% confidence interval
  StdDev    : Standard deviation of all measurements
  Gen0      : GC Generation 0 collects per 1000 operations
  Allocated : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
  1 ns      : 1 Nanosecond (0.000000001 sec)
```

src/Cask/Cask.cs
```diff
@@ -42,7 +43,7 @@ public static bool IsCask(ReadOnlySpan<char> key)
         }
 
         int lengthInBytes = Base64CharsToBytes(key.Length);
-        Debug.Assert(lengthInBytes <= MaxKeyLengthInBytes);
+        Debug.Assert(lengthInBytes <= MaxStackAlloc);
```
michaelcfanning, line 46: >MaxStackAlloc [](http://example.com/codeflow?start=38&length=13)

I am nit-picking your name but the real spirit of this limit relates to the length constraint of our keys, not a general limit in a stack allocation, yes? (I will now look at the rest of your change though maybe I'm missing something subtle).

src/Cask/Cask.cs
```diff
@@ -198,9 +200,26 @@ public static CaskKey GenerateKey(string providerSignature,
...
 
+    public static CaskKey GenerateHash(string derivationInput, CaskKey secret, int secretEntropyInBytes)
+    {
+        return GenerateHash(derivationInput.AsSpan(), secret, secretEntropyInBytes);
+    }
+
+    public static CaskKey GenerateHash(ReadOnlySpan<char> derivationInput, CaskKey secret, int secretEntropyInBytes)
+    {
+        int byteCount = Encoding.UTF8.GetByteCount(derivationInput);
+        ThrowIfDerivationInputTooLong(derivationInput, byteCount);
```
michaelcfanning, line 211: >ThrowIfDerivationInputTooLong [](http://example.com/codeflow?start=8&length=29)

what's our constraint here? i will keep my eyes open. we hadn't thought to try to limit the derivation input provided by users, to date this is left up to users.

this is something under the service provider's control, i.e., if they propose to process lots of data, well, they will assume those costs. so not sure how important it is for us to be involved.

in practice, i doubt people hash large amounts of data...

src/Cask/Cask.cs
michaelcfanning: OK, looking elsewhere I see the answer: you enforce this limit for data size other than literal key lengths. Got it.

It's not important. But it troubles me if we're not enforcing the literal limit on the key length here, that's what seems best. If values match, perhaps we could simply create an alias between two constants. If the key length we enforce differs from the max stack allocation, what we could do is supplement the code (somewhere appropriate) with an assert the ensures that MaxKeyLengthInBytes <= MaxStackAlloc.

src/Cask/Cask.cs
```diff
@@ -198,9 +200,26 @@ public static CaskKey GenerateKey(string providerSignature,
...
+        Debug.Assert(byteCount <= MaxStackAlloc);
+        Span<byte> span = stackalloc byte[byteCount];
+        Encoding.UTF8.GetBytes(derivationInput, span);
+        return GenerateHash(span, secret, secretEntropyInBytes);
+    }
+
     public static CaskKey GenerateHash(ReadOnlySpan<byte> derivationInput, CaskKey secret, int secretEntropyInBytes)
     {
+        ThrowIfDerivationInputTooLong(derivationInput);
         int hashLengthInBytes = GetHashLengthInBytes(secret, ref secretEntropyInBytes, out int providerDataLengthInBytes);
```
michaelcfanning, line 221: It's been a long week. I seem to be balking on minor points. But it troubles me that we are passing in a `CaskKey` instance for a computation that is exclusively driven by a single computed value/property of the cask key, which is simply its length in bytes.

Now, if we simply turn this into an argument, `int keyLengthInBytes`, then it would be nice if a CaskKey itself could return this value via helper/property.

I will go look more closely at CaskKey before commenting further.

src/Cask/Cask.cs
```diff
@@ -198,9 +200,26 @@ public static CaskKey GenerateKey(string providerSignature,
...
+    {
+        int byteCount = Encoding.UTF8.GetByteCount(derivationInput);
+        ThrowIfDerivationInputTooLong(derivationInput, byteCount);
+        Debug.Assert(byteCount <= MaxStackAlloc);
+        Span<byte> span = stackalloc byte[byteCount];
+        Encoding.UTF8.GetBytes(derivationInput, span);
+        return GenerateHash(span, secret, secretEntropyInBytes);
+    }
+
     public static CaskKey GenerateHash(ReadOnlySpan<byte> derivationInput, CaskKey secret, int secretEntropyInBytes)
```
michaelcfanning, line 218: >int secretEntropyInBytes [](http://example.com/codeflow?start=91&length=24)

the decoupling between the argument to specify entropy and our requirement to round this value up to a number divisible by 3 continues to trouble me.

It's weird if someone *must* provide 18 bytes (not 16) for a 128 bit key. It's also a bit weird if we possibly leave a footgun because the generated key has space for 18 bytes, while the user has retained their specified argument of 16. If our `CaskKey` object in particular is performant and useful, that data could be uniformly accessed for everything post-creation.

Another idea, just a thought, is to have an enum for the entropy. Naming would be fun for this approach.

public enum KeySize { 
None = 0,
Pwned = None,
OneHundredTwentyEightBits,
ThirtyTwoBytes,
TwoToTheNinthPower
};

src/Cask/CaskKey.cs
```diff
@@ -67,27 +76,27 @@ public static bool TryEncode(ReadOnlySpan<byte> bytes, out CaskKey key)
...
 
-    public static CaskKey Parse(string text)
+    public static CaskKey Create(string text)
     {
-        if (!TryParse(text, out CaskKey key))
+        if (!TryCreate(text, out CaskKey key))
         {
             ThrowFormat();
         }
         return key;
```
michaelcfanning, line 85: >return [](http://example.com/codeflow?start=8&length=6)

NIT: elsewhere in the code you are careful to put a blank line after closing a conditional but you've missed this one and in the other overload.

src/Cask/Cask.cs
```diff
@@ -198,9 +200,26 @@ public static CaskKey GenerateKey(string providerSignature,
...
     {
+        ThrowIfDerivationInputTooLong(derivationInput);
         int hashLengthInBytes = GetHashLengthInBytes(secret, ref secretEntropyInBytes, out int providerDataLengthInBytes);
+        Debug.Assert(hashLengthInBytes <= MaxStackAlloc);
```
michaelcfanning, line 222: >MaxStackAlloc [](http://example.com/codeflow?start=42&length=13)

You declare a limit `MaxDerivationInputLengthInBytes` that is an alias to `MaxStackAlloc` and I think that's clearer (also for the key length).

src/Cask/Limits.cs
```diff
@@ -55,4 +55,20 @@ public static class Limits
...
+
+    /// <summary>
+    /// The maximum amount of bytes that the implementation will stackalloc.
+    /// </summary>
+    /// <remarks>
+    /// Internal and constant because it is an implementation detail. All byte
+    /// limits must be less than or equal to this so that the implementation can
+    /// stackalloc unconditionally.
+    /// </remarks>
+    internal const int MaxStackAlloc = 256;
```
michaelcfanning, line 73: >256 [](http://example.com/codeflow?start=39&length=3)

alright, this very minor topic is coming into focus. you could set this value to the max key length size.

or instead you could find a place in an initialization path that ensures that all our limits stay within your `MaxStackAlloc` constraint. And then all length checks in the code itself could refer to the length limits for the actual data being processed.

src/Cask/Limits.cs
```diff
@@ -55,4 +55,20 @@ public static class Limits
     /// The maximum length of a Cask key in characters.
     /// </summary>
     public static int MaxKeyLengthInChars { get; } = BytesToBase64Chars(MaxKeyLengthInBytes);
+
+    /// <summary>
+    /// The maximum length of derivation input in bytes.
+    /// This refers to the length when UTF8-encoded for derivation input supplied as text.
+    /// </summary>
+    public static int MaxDerivationInputLengthInBytes { get; } = MaxStackAlloc;
```
michaelcfanning, line 63: >MaxStackAlloc [](http://example.com/codeflow?start=65&length=13)

I'm actually not 100% sure we can live with a 256 char limit.

src/Tests/Cask.Benchmarks/Program.cs
```diff
@@ -1,20 +1,51 @@
...
- * 
- * dotnet run -c Release --framework net8.0 Cask.Benchmarks.csproj
- */
-Runner.Run([
-    typeof(IsCaskBenchmarks),
-    typeof(CompareHashedSignatureBenchmarks),
- ]);
+if (debug)
+{
+    DebugBenchmarkRunner.Run();
```
michaelcfanning, line 43: >Run [](http://example.com/codeflow?start=25&length=3)

We could allow users to pass a specific benchmark name as an argument for debug purposes.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -267,4 +268,38 @@ private void IsCaskValidate(string key)
...
+        Assert.True(result, $"'CompareHash' failed when mock time advanced.");
+    }
+
+    [Fact]
+    public void CaskSecrets_CompareHash_UnequalLength()
+    {
+        using Mock mock = Cask.MockFillRandom(_ => { /* Leave zeros */ });
+        byte[] derivationInput = Encoding.UTF8.GetBytes("DERIVATION_INPUT");
+        string secret1 = Cask.GenerateKey(providerSignature: "TEST", allocatorCode: "88");
+        string secret2 = Cask.GenerateKey(providerSignature: "TEST", allocatorCode: "88", providerData: "OPTIONAL");
```
michaelcfanning, line 289: >providerData: "OPTIONAL" [](http://example.com/codeflow?start=90&length=24)

here's another nit, the use of optional data here is irrelevant. two discrete calls to generate key with identical params should produce unique secrets.

i feel like qualifying the second call here with a different arg is a bit of a red herring.

now, if we preserved the randomized bytes from an initial key allocation and then drove construction of second key providing the optional data, this would be a different test, a pedantic and arguably entirely unnecessary test. the keys would differ because the checksums differ. the secrets would differ because our spec says the key as a whole is the secret.

we do not have helpers that allow for this kind of finagling and I'm not sure it's useful. my suggestion here is just to zap the provider data arg because it doesn't add anything additional and may misdirect folks.

src/Cask/Cask.cs
michaelcfanning: OK, and having finally made it to your tests, I see we have unit tests. :) haha. so that means you can simply take my note, perhaps, and to always prefer the check against the limit constant relevant to the current operation.

And we probably don't need runtime asserts, though they wouldn't hurt tucked away in some initialization corner either.

src/Tests/Cask.Tests/CaskSecretsTests.cs
nguerrera: I'm trying to hit a branch that's only taken when the hashes have different lengths. The OPTIONAL makes it longer.

src/Cask/Cask.cs
nguerrera: My care factor isn't that high, but my thinking is that eyebrows (rightfully) raise around stack allocs so seeing assert against MaxStackAlloc is more comforting next to each and every stackalloc. Someone may feel more urge to go check MaxKeyLengthInBytes if that's what's near the stackalloc.

src/Cask/Cask.cs
nguerrera: I wasn't sure about this, but all the examples I saw had tiny amounts of data and it will be the only place where we cannot stackalloc unconditionally. But I'll remove this limitation.

src/Cask/Cask.cs
nguerrera: I'll change it.

src/Cask/Cask.cs
nguerrera: I'm not sure I follow everything here. This API is exactly the reason why I want the key to record how much entropy it has. It's really weird to pass (key, entropy), but it's impossible to get the information from a key in the current format.

src/Tests/Cask.Benchmarks/Program.cs
nguerrera: I thought about that, but this change is getting big and I only changed this as much as was useful to get through what I was doing. It's a good idea! There's also some ways to tell BenchmarkDotNet to let you debug and to tell it to do fewer iterations. I feel like that's what I should really be using, but the config API is quite a beast, and I just didn't have time to study this yet.

src/Tests/Cask.Tests/CaskSecretsTests.cs
nguerrera: This branch is taken before anything else is compared. I am guilty of just trying to hit the block in code coverage. I'm just going to delete this test for now.

src/Cask/Limits.cs
nguerrera: OK, I'll remove this.

src/Cask/Cask.cs
nguerrera: Since this is unchanged code, I'm leaving this and we can discuss it separately.

src/Cask/Cask.cs
nguerrera: Leaving this since it's unchanged here, a known thing we need to address that's discussed in Questions.md and we discussed how to change the format offline to fix this. In the new year, I will do that.

src/Cask/CaskKey.cs
nguerrera: Fixed.

src/Cask/Cask.cs
nguerrera: Fixed.

src/Cask/Cask.cs
nguerrera: Fixed.

src/Cask/Cask.cs
nguerrera: Fixed.

nguerrera:
I dreamt of a bug and now I'm inclined to revert the polyfills to simpler code.

Fixed in https://github.com/microsoft/cask/pull/18/commits/2e176bdf9334dcb0d842316cdcd6b07c897b9e41

But I also put no-merge on this while we consider if this is worth it.

nguerrera:
OK, I've simplified to reduce bug risk, and measured to ensure that the things I simplified weren't worth it. See updated comment with benchmarks.

michaelcfanning (approved):
:shipit:

PR #21, opened 2025-01-08
Add C3ID implementation

nguerrera:
This implements C3ID computation that we will need soon. 

The implementation differs from the one in https://github.com/microsoft/security-utilities and this seeks to become the new standard.

Changes:
- There is no use of "Microsoft"
- There is no byte array to hex conversion in the intermediate computation
- Optimized to avoid allocations
- Raw value is truncated to 12 bytes instead of 15
- "C3ID" is prefixed to the canonical representation in base64

Tests include a naive, unoptimized implementation that is easier to understand, and we check that both implementations match. We also check that we produce the same deterministic value for a sampling of hard-coded test input.

This change also fixes edge cases with empty span input in encoding polyfills.


nguerrera:
#22 should fix the build issue. Let's get that reviewed and merged first.

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    /// <summary>
+    /// Computes the Cross-Company Correlating Id (aka C3ID) bytes for the given
+    /// company and text and writes them to the destination span.
+    /// </summary>
+    public static void Compute(string company, string text, Span<byte> destination)
+    {
+        Debug.Assert(destination.Length >= SizeInBytes);
+
+        // Input: $"Cross{company}CorrelatingId:{SHA256Hex(text))}" encoded in UTF-8
```
michaelcfanning, line 32: >company [](http://example.com/codeflow?start=26&length=7)

Hm. If the input data meets the requirements specified by c3id, then that input is utterly unique in the world and can and should be hashed using an identifier that places it in the same domain as every other organization's hashing scheme. 

OTOH, specifying a per-company identifier here allows for some degree of data isolation if that's what's appropriate.

michaelcfanning, line 32: >encoded [](http://example.com/codeflow?start=68&length=7)

NIT: because we require trailing periods in doc comments, I suggest we also pick up this habit in inlined comments. i.e., prefer fully formed sentences with closing periods.

The reasoning is simple consistency and also some theoretical convenience if useful code comments get hoisted into more official doc comments.

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,96 @@
...
+
+using System.Diagnostics;
+using System.Security.Cryptography;
+using System.Text;
+
+using static CommonAnnotatedSecurityKeys.Limits;
+
+namespace CommonAnnotatedSecurityKeys;
+
+internal static class CrossCompanyCorrelatingId
```
michaelcfanning, line 12: >CrossCompanyCorrelatingId [](http://example.com/codeflow?start=22&length=25)

Just a thought, but I tend to think that a nice summary element on every class type is a nice baseline requirement, even if projects don't take on the costs of authoring nicely rendered per-method/property comments as you've done here.

This class declaration would be a nice place to describe a c3id in general, to link to a reference doc, etc. 

No action for you, I can take an action to fill this out later.

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// </summary>
+    public const int SizeInBytes = 15;
+
+    private static ReadOnlySpan<byte> CompanyPrefix => "Cross"u8;
+    private static ReadOnlySpan<byte> CompanySuffix => "CorrelatingId:"u8;
+    private static ReadOnlySpan<byte> Hex => "0123456789ABCDEF"u8;
+    private const int HexCharsPerByte = 2;
+
+    /// <summary>
+    /// Computes the Cross-Company Correlating Id (aka C3ID) bytes for the given
```
michaelcfanning, line 25: >aka [](http://example.com/codeflow?start=51&length=3)

standard convention in documentation would permit you to elide the 'aka' here. The convention below is common and, in fact, formally specified in some of our internal doc guidelines. The guideline states that in a first rendering you prefer the full name, followed by the acronym in parens, after which you can utilize the FFN for any textual references.

Fully Formed Name (FFN)

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// company and text and writes them to the destination span.
+    /// </summary>
+    public static void Compute(string company, string text, Span<byte> destination)
+    {
+        Debug.Assert(destination.Length >= SizeInBytes);
```
michaelcfanning, line 30: >Assert [](http://example.com/codeflow?start=14&length=6)

Since this is a public method, shouldn't we have formal enforcement of this argument invariant?

And also some validation for company/text?

nguerrera: It's currently an internal type. I wasn't sure if we should make this public, but probably we will want to.

nguerrera: I'll go ahead and add the validation.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+public class CrossCompanyCorrelatingIdTests
+{
+    [Theory]
+    [InlineData("", "EZ3GxRsKq+Dp21GvyCpQ")]
+    [InlineData("Hello world", "R8ogeP7QfTFvL5qAATry")]
+    [InlineData("😁", "f/BTV0j6A8km4KDw7aJz")]
+    public void Test_Basic(string text, string expected)
+    {
+        string actual = ComputeC3IDBase64(company: "Microsoft", text);
```
michaelcfanning, line 16: Hey is the proper rendering `C3ID` or `C3id`? 

Haha. It is a true fact that I treat this as a valid question despite the obvious triviality of the concern.

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,96 @@
...
+    private static ReadOnlySpan<byte> CompanyPrefix => "Cross"u8;
+    private static ReadOnlySpan<byte> CompanySuffix => "CorrelatingId:"u8;
+    private static ReadOnlySpan<byte> Hex => "0123456789ABCDEF"u8;
+    private const int HexCharsPerByte = 2;
+
+    /// <summary>
+    /// Computes the Cross-Company Correlating Id (aka C3ID) bytes for the given
+    /// company and text and writes them to the destination span.
+    /// </summary>
+    public static void Compute(string company, string text, Span<byte> destination)
```
michaelcfanning, line 28: >Compute [](http://example.com/codeflow?start=23&length=7)

OK, so here we are at the same conversation we have with CASK generally. It is the consistent expression of a c3id that provides its utility for connecting an eco-system of allocated and/or exposed secrets.

Our convention is to render this data as standard base64, as 20 encoded characters. Because we don't have any obvious need to locate these patterns in source code, etc., we don't require URL safe encoding (a useful loosening since standard base64 encoding the most highly available version as API).

Arguably, if we followed our CASK conventions, what we would do here is to have a Compute API that would accept a Span<char> for the base64 encoded result.

Certainly the doc comment should note that convention dictates expressing the output as a base64 encoded string. This would be the convention, for example, if a company dumped a list of well known compromised c3ids, etc. 

We could consider having two helpers, one that writes to Span<byte> the other to Span<string>. Perhaps co-locating these would also help make it clear that base64 is the preferred expression for this data.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
michaelcfanning: I *think* I prefer `C3id` but need I mention I have no desire to die on this hill? :)

src/Cask/CrossCompanyCorrelatingId.cs
michaelcfanning, line 12: So this class is internal for now while we play around with it?

src/Cask/CrossCompanyCorrelatingId.cs
michaelcfanning, line 28: >Compute [](http://example.com/codeflow?start=23&length=7)

Eventually we will need to think about how/whether we codify API that's centered around a specific company such as Microsoft.

I mention this in case projecting ahead to the need for an API such as `CrossCompanyCorrelatingId.ComputeMicrosoftId` or whatever makes us think differently about the core/preliminary API names we commit to.

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: I've put a summary, but it will need more work. I will work on a proper specification and update the doc comments. I'll do that separately so that I can consume this in another branch, get out that PR, then work on the spec and docs in parallel with the review for that larger PR.

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: It's now public and validating.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
nguerrera: It does not appear in any identifier other than in tests. As such, my care factor is vanishingly small. I'm just as picky about these things in public API, so no judgment. 😁 I'll happily change it, but I still don't which to use...

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: We now have a co-located helper that's string in, string out, and produces the canonical string value.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
```diff
@@ -0,0 +1,96 @@
...
+              reference: {reference}
+                 actual: {actual}
+            """);
+
+        return actual;
+    }
+
+    /// <summary>
+    /// A trival reference implementation of C3ID that is easy to understand,
+    /// but not optimized for perfomance. We compare this to the production
```
nguerrera, line 76:
```suggestion
    /// but not optimized for performance. We compare this to the production
```

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Text;
+
+using Xunit;
+
+namespace CommonAnnotatedSecurityKeys.Tests;
+
+public class CrossCompanyCorrelatingIdTests
+{
+    [Theory]
+    [InlineData("", "C3IDKZnmJO6my5hknQ5W")]
```
michaelcfanning, line 14: >C3IDKZnmJO6my5hknQ5W [](http://example.com/codeflow?start=21&length=20)

this value could become very familiar to folks troubleshooting data issues. :) i think your idea about possibly putting some exceptions/data guards in place (for the text that we hash) could make. certainly failing a `string.IsNullOrWhitespace` check would qualify.

the problem is that it can be unsatisfying to add an incomplete set of guards in circumstances. but this one may be worth adding!

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    [Fact]
+    public void C3ID_LargeText()
+    {
+        string actual = ComputeC3ID(text: new string('x', 300));
+        Assert.Equal("C3IDs+pSKJ1FmRW+7EZk", actual);
+    }
+
+    [Fact]
+    public void C3ID_NullInput_Throws()
```
michaelcfanning, line 33: >C3ID_NullInput_Throws [](http://example.com/codeflow?start=16&length=21)

how about strengthening this a bit, null, empty or all whitespace throws?

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,82 @@
...
+
+/// <summary>
+/// Cross-Company Correlating Id (C3ID) a 12-byte value used to correlate a
+/// high-entropy keys with other data. The canonical textual representation is
+/// base64 encoded and prefixed with "C3ID".
+/// </summary>
+public static class CrossCompanyCorrelatingId
+{
+    /// <summary>
+    /// The size of a Cross-Company Correlating ID (aka C3ID) in raw bytes.
```
michaelcfanning, line 19: >aka [](http://example.com/codeflow?start=52&length=3)

nit: remove `aka `

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// </summary>
+    public const int RawSizeInBytes = 12;
+
+    private static ReadOnlySpan<byte> Prefix => "C3ID"u8;
+    private static ReadOnlySpan<byte> PrefixBase64Decoded => [0x0B, 0x72, 0x03];
```
michaelcfanning, line 24: >PrefixBase64Decoded [](http://example.com/codeflow?start=38&length=19)

add comment communicating that this is the bytewise encoding of the base64 encoded literal `c3id`?

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    /// <summary>
+    /// Computes the C3ID for the given text in canonical textual form.
+    /// </summary>
+    public static string Compute(string text)
+    {
+        ThrowIfNull(text);
```
michaelcfanning, line 31: >ThrowIfNull [](http://example.com/codeflow?start=8&length=11)

`ThrowIfNullEmptyOrWhitespace` and make it so?

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,82 @@
...
+        PrefixBase64Decoded.CopyTo(bytes);
+        Compute(text, bytes[PrefixBase64Decoded.Length..]);
+        return Convert.ToBase64String(bytes);
+    }
+
+    /// <summary>
+    /// Computes the raw C3ID bytes for the given text and writes them to the
+    /// destination span.
+    /// </summary>
+    public static void Compute(string text, Span<byte> destination)
```
michaelcfanning, line 43: >Compute [](http://example.com/codeflow?start=23&length=7)

You know, I've wondered, why do some API go ahead and return a span in some API and in others we provide an argument.

And you've answered it! If you would otherwise have an overload that would differ only by return type, you are stuck!

honestly, this is an (unavoidable as far as I can see) idiosyncrasy that's a bit regrettable. 

but what can you do? the fact is that string is now effectively a replaceable entity and the only way to provide functionality that diverges is by unique parameterization.

another bit of weirdness is that in places where you *can* control the return value and use Span the ReadOnly version might be appropriate (and easy to forget to use??)

I'm going on about this too much, forgive me, it's fun to talk to you about this stuff, but I'm wondering about alternate API shapes, e.g.:

void Compute(string text, out string destination);
void Compute(string text, Span<byte> destination);

also, potentially zapping exception raising by simply providing these:

bool TryCompute(string text, out string destination);
bool TryCompute(string text, Span<byte> destination);

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,82 @@
...
+        Compute(text.AsSpan(), destination);
+    }
+
+    /// <summary>
+    /// Computes the raw C3ID bytes for the given UTF-16 encoded text sequence
+    /// and writes them to the destination span.
+    /// </summary>
+    public static void Compute(ReadOnlySpan<char> text, Span<byte> destination)
+    {
+        ThrowIfDestinationTooSmall(destination, RawSizeInBytes);
```
michaelcfanning, line 55: >ThrowIfDestinationTooSmall [](http://example.com/codeflow?start=8&length=26)

Ah, if we accept my idea around preventing empty or whitespace we'd want to add it here. So that the upstream doesn't duplicate the work.

I'm on the fence. On the one hand, it's perfectly legal to generate a hash for one or more tabs/spaces, etc. on the other hand, this data will *never* comprise something we should generate a c3id for. and certainly hashes of the empty string are a notorious source of data noise.

Your thoughts?

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// </summary>
+    private static class ReferenceCrossCompanyCorrelatingId
+    {
+        public static string Compute(string text)
+        {
+            // Compute the SHA-256 hash of the UTF8-encoded text
+            Span<byte> hash = SHA256.HashData(Encoding.UTF8.GetBytes(text));
+
+            // Prefix the result with "C3ID" UTF-8 bytes and hash again
+            hash = SHA256.HashData([.. "C3ID"u8, .. hash]);
```
michaelcfanning, line 87: >HashData [](http://example.com/codeflow?start=26&length=8)

Wait, aren't we supposed to be using the `PrefixBase64Decoded` const bytes here?

michaelcfanning: Ah, I see all tests are passing because your optimized version does the same thing. I thought we were going to use the decoded bytes directly not the UTF8 bytes of the encoded form.

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: > I've wondered, why do some API go ahead and return a span in some API and in others we provide an argument.

> And you've answered it! If you would otherwise have an overload that would differ only by return type, you are stuck!

That's not the answer! 😁

The point of `ProduceData(..., Span<byte> destination)` is that the caller can tell the API where to write the data. If the caller wants to write it in the middle of an array somewhere, they can do that without allocating and copying. Or, if they know it will be a small amount of data, they can put it on the stack and avoid all allocations.

`Span<byte> ProduceNewData(...)` does not make sense. Where does it put that new data? It can't be on the stack, and if it returns a span comprising an entire new byte array on the heap, then nothing has been gained vs `byte[] ProduceNewData(...)`.

If an API doesn't produce new data, but returns a slice of existing data, then you can return a span and profit. For example, `Span<T> Slice(int start, int length)`.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
nguerrera: > Wait, aren't we supposed to be using the PrefixBase64Decoded const bytes here?

I misunderstood if that's what you wanted. However, it does seem valuable to me that I can specify C3ID and write it as above without having to mention magic numbers or make additional Base64 conversions. The PrefixBase64Decoded in the production implementation is then merely a trick to have fewer allocations.

The fact that I misunderstood also makes me think it would be better to stop trying to use the same string for both purposes. The prefix on the output and the prefix to the SHA input can be arbitrarily different and maybe it would more clearly communicate that this is the case if they were completely different.

src/Cask/CrossCompanyCorrelatingId.cs
michaelcfanning: And so with my corrected optics, this kind of pattern actually could make sense, because we're effectively handling the return value in a consistent way (from an API shape perspective).

I don't see a real downside to going this way. There's arguably more complexity at the call site. Thoughts?

bool TryCompute(string text, out string destination);
bool TryCompute(string text, Span<byte> destination);

michaelcfanning: Hm did my code flow reply get lost? If so, then duh, of course, this is the way for the caller to provide the destination.

With the duh corrected, my comment on using this consistent pattern is worth considering. In my other comment (which could appear at any second, maybe) I note this arguably creates more call site complexity. 

bool TryCompute(string text, out string destination);
bool TryCompute(string text, Span<blah> destination);

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: Will do.

Incidentally, completely separate, but you may wonder why this span return is OK. It looks like it allocates a new array on every property call, after all! It turns out that this particular pattern is optimized by the C# compiler. The data is actually burned into the PE file and the span points straight to it using an RVA static field.

I wish there were a unique syntax for this. It would make it clearer that magic is happening. There was a language proposal, but it didn't go anywhere. https://github.com/dotnet/csharplang/issues/5295

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: I went to do it and the most annoying thing is validating whitespace on the UTF-8 overload. I thought we might want this to be public since I could imagine something processing a ton of UTF-8 data and wanting to produce C3IDs without allocation and UTF-8 -> UTF-16 -> UTF-8 conversion.

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: Actually, it isn't so magic anymore with collection expressions and no `new byte[]` that doesn't actually allocate an array. 

Collection expressions are documented to optimize:

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/collection-expressions

> The compiler uses static analysis to determine the most performant way to create the collection declared with a collection expression.

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: > I note this arguably creates more call site complexity.

I agree and I think this isn't worth it. There should be an articulated reason other than lining up with span alternatives to have an out param. The non-span convenience alternatives should be, well, convenient. There's plenty of BCL prior art for convenient array/string return and fast/flexible span parameter. See, for example, `SHA256.HashData` where one overload returns `byte[]` and another writes to a `Span<byte>`.

Anoth thing about this case, the "raw" bytes and string are sufficiently different (no prefix, no base64 in raw form), that I have now decided to have separate `Compute` and `ComputeRaw`. I think it's worth it since we want to lead nearly everyone to non-raw and anyone who uses raw should understand the implications. Furthermore, it's possible we will want overloads to write non-raw to span, (or less likely, return raw as byte[]) and this leaves room for that.

src/Cask/CrossCompanyCorrelatingId.cs
nguerrera: Since this is not going to catch all bad input, I am going to compromise and only find ASCII-range whitespace.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
nguerrera: I'll go ahead and use the decoded prefix in both places.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
```diff
@@ -0,0 +1,129 @@
...
+
+              reference: {reference}
+                 actual: {actual}
+            """);
+
+        return actual;
+    }
+
+    /// <summary>
+    /// A trival reference implementation of C3ID that is easy to understand,
```
nguerrera, line 106:
```suggestion
    /// A trivial reference implementation of C3ID that is easy to understand,
```

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
nguerrera: We decided offline to revert back to what I had in the end.

src/Tests/Cask.Tests/CrossCompanyCorrelatingIdTests.cs
nguerrera: I went with `C3Id` now. I need it in non-test code in my next change. Still not public, but I think it's worth settling this right now and not having noise of renaming it mixed in with other changes.

My reason for `C3Id` over `C3id` is a nod to my much younger self that rewrote the FxCop casing tokenizer ~20 years ago. That would parse this as two tokens C3 and id. And only `C3Id` respects the guidelines with that tokenization. 😁

I'd ask to get rid of the number in the middle of an acronym, but I think I've caused enough trouble asking for algo changes.

src/Cask/CrossCompanyCorrelatingId.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        ThrowIfEmpty(textUtf8);
+        ThrowIfDestinationTooSmall(destination, RawSizeInBytes);
+
+        // Produce input for second hash: "C3ID"u8 + SHA256(text)
+        Span<byte> input = stackalloc byte[Prefix.Length + SHA256.HashSizeInBytes];
+        Prefix.CopyTo(input);
+        SHA256.HashData(textUtf8, input[Prefix.Length..]);
+
+        // Perform second hash, truncate, and copy to destination.
```
nguerrera, line 87: I'm getting concerned about the cost here. Two rounds of SHA256 is a lot of compute for a 12 byte hash. 

For context from offline conversation: the rationale (as I understand it, but I may not) for two rounds in C3ID is to make it possible to compute the C3ID from the plain SHA256. Without sharing the key and without adding C3ID to their toolkit, someone can send you the SHA256 of a key, and then you can convert that to C3ID.

We should maybe reconsider if this scenario is worth the cost added to CASK scenarios. I'll provide some data on that cost in a minute.

nguerrera: ## Without C3ID in hash scenarios
From https://github.com/microsoft/cask/pull/18#issuecomment-2555778281

| Method             | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------------- |-----------:|----------:|----------:|-------:|----------:|
| CompareHash_Cask   | 451.609 ns | 3.9292 ns | 3.6753 ns |      - |         - |
| CompareHash_Floor  | 294.875 ns | 2.7652 ns | 2.5866 ns |      - |         - |
| GenerateHash_Cask  | 300.978 ns | 1.3669 ns | 1.2786 ns | 0.0091 |     176 B |
| GenerateHash_Floor | 225.569 ns | 1.6476 ns | 1.2864 ns | 0.0057 |     112 B |

## With C3ID in hash scenarios
From my work-in-progress branch:

| Method                         | Mean     | Error   | StdDev  | Gen0   | Allocated |
|------------------------------- |---------:|--------:|--------:|-------:|----------:|
| CompareHash_Cask               | 856.6 ns | 6.06 ns | 5.37 ns |      - |         - |
| CompareHash_Floor              | 298.5 ns | 3.69 ns | 3.27 ns |      - |         - |
| GenerateHash_Cask              | 725.1 ns | 5.20 ns | 4.86 ns | 0.0114 |     216 B |
| GenerateHash_Floor             | 226.3 ns | 0.78 ns | 0.69 ns | 0.0057 |     112 B |

There are some other changes in this branch so this doesn't quite isolate the C3ID cost. However, I also measured on this machine that a single SHA256 of Cask key sized data -- without allocation, copying, or encoding conversion -- takes about 120ns on this machine. That's a lot relative to the "floor" scenarios we're aiming to replace.

Up to now, I was confident in being able to tell a good story about how the extra Cask costs were a fair price to pay for Cask features, but I am not sure C3ID is living up to that now.

nguerrera: With that said, I think it would be good to take this change and discuss options to improve this in follow-up.

michaelcfanning (approved):
:shipit:

src/Cask/CrossCompanyCorrelatingId.cs
michaelcfanning: Ack on this and let's talk more.

PR #22, opened 2025-01-08
Do not try to run net472 tests on non-Windows using Mono

nguerrera:
The PR validation builds started failing when Mono was removed from the latest images.

We never intended to support Mono. Non-Windows consumers must use .NET 8+.

This change also cleans up and shares some helper flags for inspecting the build environment


src/Directory.Build.props
```diff
@@ -51,5 +51,14 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
+  <PropertyGroup Label="Build Environment Helpers">
+    <_OSIsWindows>false</_OSIsWindows>
+    <_OSIsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</_OSIsWindows>
+    <_OSIsX64>false</_OSIsX64>
+    <_OSIsX64 Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">true</_OSIsX64>
+    <_MSBuildIsNETFramework Condition="'$(MSBuildRuntimeType)' == 'full'">true</_MSBuildIsNETFramework>
+    <_MSBuildIsNETFramework Condition="'$(_MSBuildIsNETFramework)' == ''">false</_MSBuildIsNETFramework>
```
nguerrera, line 60: Inconsistency here, fixing

src/Directory.Build.props
```diff
@@ -51,5 +51,14 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
+  <PropertyGroup Label="Build Environment Helpers">
+    <_OSIsWindows>false</_OSIsWindows>
```
michaelcfanning, line 55: >_OSIsWindows [](http://example.com/codeflow?start=5&length=12)

Just curious, do you use the convention of an underscore prefix here it make it very obvious elsewhere these are custom defined vars not built-ins?

src/Directory.Build.props
```diff
@@ -51,5 +51,14 @@
...
+    <_OSIsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</_OSIsWindows>
+    <_OSIsX64>false</_OSIsX64>
+    <_OSIsX64 Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'X64'">true</_OSIsX64>
+    <_MSBuildIsNETFramework>false</_MSBuildIsNETFramework>
+    <_MSBuildIsNETFramework Condition="'$(MSBuildRuntimeType)' == 'full'">true</_MSBuildIsNETFramework>
```
michaelcfanning, line 60: >_MSBuildIsNETFramework [](http://example.com/codeflow?start=5&length=22)

Just curious, there's no mechanism for forcing predicate eval to populate the contents of an element?

So that we basically have a single element that evals to either true/false.

<MSBuildIsNETFramework>$(MSBuildRuntimeType)' == 'full'</_MSBuildIsNETFramework>

If not, I understand why you first explicitly initialize to false and then eval these conditions.

src/Tests/Cask.Tests/Cask.Tests.csproj
```diff
@@ -1,21 +1,21 @@
...
+    <_BuildCpp Condition="$(_OSIsWindows) and $(_MSBuildIsNETFramework) and $(_OSIsX64)">true</_BuildCpp>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\Cask\Cask.csproj" />
     <None Update="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(BuildCpp)' == 'true'">
+  <ItemGroup Condition="$(_BuildCpp)">
```
michaelcfanning, line 18: >_BuildCpp [](http://example.com/codeflow?start=26&length=9)

yes! here's the nice benefit of your approach, no direct comparisons to the explicitly set `true` value.

Still, I wonder, if _BuildCpp isn't set at all, won't `Condition` interpret that as false? It seems unlikely it could possibly interpret an non-existent value as true.

Does this circumstance error out the build?

TLDR: if you delete the explicit initialization above of _BuildCpp to false, what happens in our build?

src/Directory.Build.props
nguerrera: There isn't a one-liner for this that I'm aware of, but I wish there were.

src/Directory.Build.props
nguerrera: Yes, that is the convention.

src/Tests/Cask.Tests/Cask.Tests.csproj
michaelcfanning: Gotcha. This may be a decent pattern. The explicit initialization elsewhere it not too burdensome and occurs only in a single place. 

Because MSBUILD is, in fact, strict on requiring explicit initialization, you can also be confident that the pattern is consciously, completely realized. 

So I guess this really just resolves to a readability issue. In C#, the syntax colorization and other editor support makes the explicit tests against boolean literal unnecessary and in many cases annoying.

Not sure what I think about the impact on readability in this XML context by eliding 'true'. But I'm game, let's give it a shot. If there's a good description name for the condition (like `_BuildCpp`) I tend to think your new pattern is preferable.

PR #26, opened 2025-01-10
Add standard BNF for 256-bit primary key.

michaelcfanning:
This change adds a spare markdown standard BNF specification for CASK keys.

The markdown additionally describes how to dereference/decompose an encoded form of a key or its bytewise (decoded) form to obtain constituent elements.

This change also adds a naive test helper that breaks down a 3-byte sequence into 4 easily obtained 6-bit values (rendered as bytes). It is not clear this code should be productized. This code is tested by a brute force unit tests that takes over 30 seconds on .NET framework to run (will be updated tomorrow to resolve this issue).


docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
...
+
+<month> ::= 'A'..'L' ; For months January to December
+
+<day> ::= 'A'..'Z' | 'a'..'f' ; 'A' = day 1, 'B' = day 2, ... 'e' = day 30, ... 'f' = day 31
+
+<hour> ::= 'A'..'X' ; Represents hours 0-23. 'A' = hour 0 (midnight), ... 'X' = hour 23.
+
+<key-type> ::= <256-bit-key> | 
+
+<256-bit-key> ::= 'A' | <256-bit-hash> | <384-bit-hash>
```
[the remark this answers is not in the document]
michaelcfanning: Oops, yes, thanks.

docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
...
+<four-zeros-prefix-base64> ::= 'A'..'D' ; Base64 characters starting with 0000b (indices 0-3).
+
+<base64-two-zeros-suffix> ::= 'A' | 'E' | 'I' | 'M' | 'Q' | 'U' | 'Y' | 'c' ; Base64 characters ending in 00b. These indices are all
+                            | 'g' | 'k' | 'o' | 's' | 'w' | '0' | '4' | '8' ; multiple of 4 (or the value of 0b), a fact that may be
+                                                                            ; useful in some contexts.
+```
+## Byte-wise Rendering
+|Byte Range|Decimal|Hex|Binary|Description|
+|-|-|-|-|-|
+|key[0] - key[31]|0 - 256|0x0 - 0xFF|00000000b - 11111111b|256 bits of random data produced by a cryptographically secure RNG|
```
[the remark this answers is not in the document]
michaelcfanning: Yes! As long as human are involved in code tinkering, I suppose we can expect one-off errors to persist.

docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
...
+|key[32]|0d|0xFF|00000000b| A reserved byte to enforce 3-byte alignment.
```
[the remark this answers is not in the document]
michaelcfanning: Right, thanks!

docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
...
+|[optional 3-bye sequences]|0 - 256|0x0 - 0xFF|00000000b - 11111111b|Provider-defined data of arbitrary form.
+|key[keyLength - 6]|0|0xFF|00000000b - 11111100b| (KeyKind)key[keyLength - 6] >> 2 (leading 6 bits comprises kind enum + two bits of version number). Currently, [keyLength - 6] & 0xFC == 0 (as these bits are undefined in the version data)
+|key[key.Length - 5]|0|0xFF|00000000b - 11110000b| (CaskVersion)([key.Length - 6] & 0xFC << 4) & key[key.Length - 5] >> 4) (leading 4 bits comprise final 4 bits of version + 4 bits of zero padding).
+|key[keyLength - 4] - key[keyLength - 1]|0 - 256|0x0 - 0xFF|00000000b - 11111111b|CRC32(key[0] - key[key.Length - 5])
```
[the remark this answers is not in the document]
michaelcfanning: Good catch and a good note re: ellipses. Thanks.

docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
+# CASK 256-bit Primary Keys
+## Standard Backus-Naur Form (BNF)
+```
+<key> ::= <payload-data> <checksum>
+
+<payload-data> ::= <random-data> <reserved> [<optional-fields>] <cask-signature> <provider-id> <timestamp> <key-type> <version>
+
+<random-data> ::= 42 * <base64url> <base64-two-zeros-suffix>; The total random data comprises 256 bits encoded as 42
```
nguerrera, line 8: nit: space before comment.

docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
...
+                                                                 ; characters x 6 bit bits of random data = 252 bits and
```
nguerrera, line 9: nit: line up semicolons.

docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
...
+                                                                 ; 1 character providing 4 bits of random data padded with 00b.
+
+<reserved> ::= '0' ; Reserved for future use.
+
+<optional-fields> ::= { <optional-field> } ; Zero or more 4-character (24 bit) sequences of optional data.
+
+<optional-field> ::= 4 * <base64url> ; Each optional field is 4 characters (24 bits). This keeps
+                                          ; data cleanly aligned along 3-byte/4-encoded character boundaries
```
nguerrera, line 17: nit: line up semicolons.

src/Cask/KeyKind.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Specifies a 256-bit primary API or other security key. The enum
+    /// value '0' comprises the base64 index that maps to character 'A'.
+    /// </summary>
+    Key256Bit = 0, // Base64 index 0 == 'A'
+
+    /// <summary>
+    /// Specifies a CASK hashed signature that incorporates a Keyed Hash
+    /// Message Authenticode of a derivation input generated by SHA-256
+    /// (HMAC-SHA-256) along with other CASK data such as the generated
+    /// C3ID of the CASK key used to initialize the HMAC instance.
```
nguerrera, line 28: Avoid referring to HMAC 'instance'. I think this is an implementation detail of legacy .NET crypto API.

src/Cask/KeyKind.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// Specifies a 256-bit primary API or other security key. The enum
+    /// value '0' comprises the base64 index that maps to character 'A'.
+    /// </summary>
+    Key256Bit = 0, // Base64 index 0 == 'A'
+
+    /// <summary>
+    /// Specifies a CASK hashed signature that incorporates a Keyed Hash
+    /// Message Authenticode of a derivation input generated by SHA-256
```
nguerrera, line 26:
```suggestion
    /// Message Authentication Code of a derivation input generated by SHA-256
```

src/Cask/KeyKind.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// (HMAC-SHA-256) along with other CASK data such as the generated
+    /// C3ID of the CASK key used to initialize the HMAC instance.
+    /// </summary>
+    Hash256Bit = 7,
+
+    /// <summary>
+    /// Specifies a CASK hashed signature that incorporates a Keyed Hash
+    /// Message Authenticode of a derivation input generated by SHA-384
```
nguerrera, line 34:
```suggestion
    /// Message Authentication Code of a derivation input generated by SHA-384
```

src/Cask/KeyKind.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// (HMAC-SHA-384) along with other CASK data such as the generated
+    /// C3ID of the CASK key used to initialize the HMAC instance.
```
nguerrera, line 36: Again, avoid 'instance'

src/Cask/ThreeByteSequence.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        if (bytes.Length != 3)
+        {
+            throw new ArgumentException("Three-byte sequence must be exactly three bytes long.", nameof(bytes));
+        }
+
+        Bytes = bytes.ToArray();
+    }
+
+    public byte[] Bytes { get; }
```
nguerrera, line 20: The backing storage here could be `int` so that the data is inlined into the struct and this bit twiddling can be done without heap allocation.

src/Cask/ThreeByteSequence.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    public string Encoded => Base64Url.EncodeToString(Bytes);
```
nguerrera, line 22: Avoid properties that allocate on every call. I suggest `string Encode()` method. The core API shouldn't use this and only allocate a single, complete key/hash string at the very end. If it's not a property, it will stick out more as an expense that can be avoided.

src/Cask/KeyKind.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace CommonAnnotatedSecurityKeys;
+
+internal enum CaskVersion : byte
+{
+    /// <summary>
+    /// Specifies version 1.0.0 of CASK.
+    /// </summary>
+    OneZeroZero,
+
+    Reserved = 0xfe,
```
nguerrera, line 13: `Reserved` should usually be avoided in enums. Is this API meant to stay internal?

We might consider putting these masks in separate constants if the enums will be made public.

docs/PrimaryKey.md
nguerrera, line 61: We had said we'd fill the padding byte with random, I think, on our last call. I am just curious, did you change your mind?

docs/PrimaryKey.md
```diff
@@ -0,0 +1,65 @@
...
+                                          ; facilitating readability of encoded form as well as byte-wise use.
+
+<cask-signature> ::= 'JQQJ' ; Fixed signature identifying the CASK key
+
+<provider-id> ::= 4 * <base64url> ; Provider identifier (24 bits)
```
nguerrera, line 22: This is called provider `providerSignature` in the API, should we change that to `providerId` or this to `provider-signature`?

docs/PrimaryKey.md
```diff
@@ -0,0 +1,83 @@
...
+
+<base64-two-zeros-suffix> ::= 'A' | 'E' | 'I' | 'M' | 'Q' | 'U' | 'Y' | 'c' ; Base64 characters ending in 00b. These indices are all
+                            | 'g' | 'k' | 'o' | 's' | 'w' | '0' | '4' | '8' ; multiple of 4 (or the value of 0b), a fact that may be
+                                                                            ; useful in some contexts.
+```
+## Byte-wise Rendering
+|Byte Range|Decimal|Hex|Binary|Description|
+|-|-|-|-|-|
+|decodedKey[0] - key[31]|0..255|0x0..0xFF|00000000b..11111111b|256 bits of random data produced by a cryptographically secure RNG|
+|decodedKey[32]|0|0xFF|00000000b| A reserved byte to enforce 3-byte alignment. This data can be repurposed by providers (particularly the final base64 encoded character).
```
nguerrera, line 61: I'd prefer not to have "this data can be repurposed" in the spec.

docs/PrimaryKey.md
```diff
@@ -0,0 +1,83 @@
...
+|[optional 3-byte sequences]|0..255|0x0..0xFF|00000000b..11111111b|Provider-defined data of arbitrary interpretation.
+|decodedKey[^15..^12]| 37, 4, 9  |0x25, 0x04, 0x09| 00100101b, 00000100b, 00001001b | Decoded 'JQQJ' signature.
+|decodedKey[^12..^9]|0..255|0x0..0xFF|00000000b..11111111b| Provider identifier, e.g. , '0x4c', '0x44', '0x93' (base64 encoded as 'TEST')
+|decodedKey[^9..^]||||Time stamp data encoded in 4 six-bit blocks for YMDH.
+|decodedKey[^6] >> 2|0, 7, or 8|0xFF|00000000b..11111100b| Leading 6 bits comprises kind enum followed by 2 bits of reserved padding. key[key.Length - 6] & 0xFC == 0.
+|decodedKey[^5] >> 4]|0|0xFF|00000000b..11110000b| Leading 4 bits comprise 4 bits of reserved version information + 4 bits of zero padding (to preserve consistent rendering of the subsequence checksum data).
+|decodedKey[..^4]|0 - 255|0x0..0xFF|00000000b..11111111b|CRC32(key[0] - key[key.Length - 5])
```
nguerrera, line 68:
```suggestion
|decodedKey[..^4]|0 - 255|0x0..0xFF|00000000b..11111111b|CRC32(key[..^4])
```

docs/PrimaryKey.md
```diff
@@ -0,0 +1,83 @@
...
+```
+## Byte-wise Rendering
+|Byte Range|Decimal|Hex|Binary|Description|
+|-|-|-|-|-|
+|decodedKey[0] - key[31]|0..255|0x0..0xFF|00000000b..11111111b|256 bits of random data produced by a cryptographically secure RNG|
+|decodedKey[32]|0|0xFF|00000000b| A reserved byte to enforce 3-byte alignment. This data can be repurposed by providers (particularly the final base64 encoded character).
+|[optional 3-byte sequences]|0..255|0x0..0xFF|00000000b..11111111b|Provider-defined data of arbitrary interpretation.
+|decodedKey[^15..^12]| 37, 4, 9  |0x25, 0x04, 0x09| 00100101b, 00000100b, 00001001b | Decoded 'JQQJ' signature.
+|decodedKey[^12..^9]|0..255|0x0..0xFF|00000000b..11111111b| Provider identifier, e.g. , '0x4c', '0x44', '0x93' (base64 encoded as 'TEST')
+|decodedKey[^9..^]||||Time stamp data encoded in 4 six-bit blocks for YMDH.
```
nguerrera, line 65:
```suggestion
|decodedKey[^9..^6]||||Time stamp data encoded in 4 six-bit blocks for YMDH.
```

docs/PrimaryKey.md
michaelcfanning: yes!

docs/PrimaryKey.md
michaelcfanning: what do you think? one reminder as we declare a formal naming discussion, this data is likely to be decomposed into two smaller components, a reserved company id and a per-company provider id. MS would likely reserve `MS` and `AZ` and require all these identifiers to be prefixed with this company reserved component. this is why i switched to `id` mentally but honestly that leap was a very tenuous one.

docs/PrimaryKey.md
```diff
@@ -57,9 +57,27 @@
...
+|-|-|-|
+|encodedKey[..42] | 'A'..'_' | 252 bits of randomized data generated by cryptographically secure RNG
+|encodedKey[42] | 'A'..'_' | 4 bits of randomized data followed by 2 bits provider reserved data.
+|encodedKey[43] | 'A'..'_' | 6 bits of provider reserved data. This character should be reserved for stable, readable information, e.g., a provider-specific key kind.
+|encodedKey[^20..^16]|'JQQJ'| Fixed CASK signature.
+|encodedKey[^16]|'A'..'Z'\|'a'..'z'| The first character of the provider signature which must be alphabetic (upper-case indicating a customer-managed as opposed to service managed secret).
+|encodedKey[^15..^12]|('A'..'Z-\_)\|('a'..'z-\_)| | The remaining three encoded characters. Any alphabetic characters must be consistently upper- or lower-case (to distinguish customer- vs. service-managed secrets).
+|encodedKey[^12]|'A'..'_'|Represents the year of key allocation, 'A' (2024) to '_' (2087)|
+|encodedKey[^13]|'A'..'L'|Represents the month of key allocation, 'A' (January) to 'L' (December)|
+|encodedKey[^14]|'A'..'Z'\|'a'..'e'|Represents the day of key allocation, 'A' (0) to 'f' (31)|
```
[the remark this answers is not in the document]
michaelcfanning: haha! good catch, i'd caught this as well. be honest: did you see it or did copilot? :) please don't ask me this question.

docs/PrimaryKey.md
```diff
@@ -57,9 +57,27 @@
...
+
+## Primary 256-bit Key Base64-Encoded Rendering
+|String Range|Text Value|Description|
+|-|-|-|
+|encodedKey[..42] | 'A'..'_' | 252 bits of randomized data generated by cryptographically secure RNG
+|encodedKey[42] | 'A'..'_' | 4 bits of randomized data followed by 2 bits provider reserved data.
+|encodedKey[43] | 'A'..'_' | 6 bits of provider reserved data. This character should be reserved for stable, readable information, e.g., a provider-specific key kind.
+|encodedKey[^20..^16]|'JQQJ'| Fixed CASK signature.
+|encodedKey[^16]|'A'..'Z'\|'a'..'z'| The first character of the provider signature which must be alphabetic (upper-case indicating a customer-managed as opposed to service managed secret).
+|encodedKey[^15..^12]|('A'..'Z-\_)\|('a'..'z-\_)| | The remaining three encoded characters. Any alphabetic characters must be consistently upper- or lower-case (to distinguish customer- vs. service-managed secrets).
```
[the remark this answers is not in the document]
michaelcfanning: yes, thanks. not sure how best to render this, using regex format which is non-obvious. lots of syntax conventions mixed in here between the C# range stuff, some BNF conventions and regex char ranges. all suggestions for making this more consistent are welcome.

docs/PrimaryKey.md
michaelcfanning: oops yes.

docs/PrimaryKey.md
michaelcfanning: let's talk about this tomorrow. we have this spare data here which allows a provider to inject a sub-key kind or some other data, possible to encode 64 values. i have observed in other applications of this technique that it is very common that this is all that a provider needs of optional data. if we give them this free space, they don't need to commit to a full 3-byte optional sequence.

note that we can play the same trick with a 16 byte or 64-byte key, each of which will carry along 2 bytes of free padding. A 384 bit key falls on a clean 3-byte boundary and there can be no cuteness for this key length: you'd need to commit to a full 3 additional bytes.

docs/PrimaryKey.md
```diff
@@ -57,9 +57,27 @@
...
-|key[key.Length - 5]|0|0xFF|00000000b - 11110000b| (CaskVersion)([key.Length - 6] & 0xFC << 4) & key[key.Length - 5] >> 4) (leading 4 bits comprise final 4 bits of version + 4 bits of zero padding).
-|key[keyLength - 4] - key[keyLength - 1]|0 - 256|0x0 - 0xFF|00000000b - 11111111b|CRC32(key[0] - key[key.Length - 5])
+|decodedKey[0] - key[31]|0..255|0x0..0xFF|00000000b..11111111b|256 bits of random data produced by a cryptographically secure RNG|
+|decodedKey[32]|0|0xFF|00000000b| A reserved byte to enforce 3-byte alignment. This data can be repurposed by providers (particularly the final base64 encoded character).
+|[optional 3-byte sequences]|0..255|0x0..0xFF|00000000b..11111111b|Provider-defined data of arbitrary interpretation.
+|decodedKey[^15..^12]| 37, 4, 9  |0x25, 0x04, 0x09| 00100101b, 00000100b, 00001001b | Decoded 'JQQJ' signature.
+|decodedKey[^12..^9]|0..255|0x0..0xFF|00000000b..11111111b| Provider identifier, e.g. , '0x4c', '0x44', '0x93' (base64 encoded as 'TEST')
```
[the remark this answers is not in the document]
michaelcfanning: Yes I tend to agree, I switched to three dots for the ranges. I wonder if we should embed the actual ellipsis character.

I note that I'm not using the C# syntax for binary literals will do that in a follow-on change.

src/Cask/KeyKind.cs
michaelcfanning: yes

src/Cask/KeyKind.cs
michaelcfanning: yes

src/Cask/KeyKind.cs
michaelcfanning: Deleted this text in preference of your concise suggestion.

docs/PrimaryKey.md
michaelcfanning: nuts, thanks.

src/Cask/KeyKind.cs
michaelcfanning: yes, next PR>

docs/PrimaryKey.md
nguerrera: Yes, `id` is better.

src/Cask/ThreeByteSequence.cs
michaelcfanning: Let's talk about this one some more, want to get your thoughts. For now, I just made us accept the cost of generating this on struct creation.

docs/PrimaryKey.md
michaelcfanning: Let's talk, relates to a necessary conversation around how to specify optional bytes. So, I was thinking, perhaps we allow for a single API that specifies your optional arguments. This data would be expressed as bytes. If you passed a single byte, and we had a free byte at the end of your random data, we'd use it. If you passed two bytes, and there were two free bytes of padding at the end of the random component (as there are for 16 and 64 byte allocations), we'd stuff your data there. In cases where the length of your optional data did not permit stuffing it in padding, and in cases where the user simply requested at least 3 bytes, we'd allocate a full 3-byte optional chunk. Any padding that might exist at the end of randomized data region for these cases could be zeroed out or could contain additional entropy.

Just thinking aloud. This is all contrary to our current API which requires specifying the optional data a string, which has advantages (in that it emphasizes we are looking for the encoded interpretability of this data). Following that model, if users provided a single character, and we had a single padding byte, we'd use that space. If the user provided a 2-char value and we had at least two bytes padding, we'd take that slot. If a user provided 3-chars, we'd pad this data with a leading 'A' and use a 3-byte sequence. For 4-byte char, a standard 3-byte sequence.

docs/PrimaryKey.md
michaelcfanning: OK, so see my complex note elsewhere on how the API could function. The behavior is easily described but is it clean/simple/desirable? Dunno. For now, I am going to take your advice and required this padding to be zeroed out unless/until we decide otherwise. I think we certainly need more clarity on our handling of optional data in the API (currently required as an optional string, which I do tend to think is a good idea).

michaelcfanning:
At this point, I've done enough tinkering that we definitely should reset and scan the end result in detail. HMAC coming next. btw - does this stuff live in devdocs? or docs?

---
In reply to: [2550418565](https://github.com/microsoft/cask/pull/26#issuecomment-2550418565) [](http://example.com/codeflow?ancestors=2550418565)

docs/PrimaryKey.md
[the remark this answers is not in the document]
michaelcfanning: Oops, yes, missed this, the `hash` stuff would only be relevant in an uber BNF. And maybe we should just have that? I think so. The only distinction between the key scenario and HMAC is that for the HMAC we have 12 bytes of c3id that immediately follow the randomized data. That's it. If we split the definitions, we'll have lots of duplication and have to make trivial fixes in two places, always a recipe for disaster.

docs/PrimaryKey.md
```diff
@@ -0,0 +1,84 @@
...
+|encodedKey[43] | 'A' | 6 bits of reserved data specified as 'A', base64 character index zero.
+|encodedKey[44..^20]|'A'...'_'|0 or more 4-character sequences comprising provider optional data, of arbitrary interpretation.
+|encodedKey[^20..^16]|'JQQJ'| Fixed CASK signature.
+|encodedKey[^16]|'A'...'Z'\|'a'...'z'| The first character of the provider signature which must be alphabetic (upper-case indicating a customer-managed as opposed to service managed secret).
+|encodedKey[^15..^12]|('A'...'Z'-\_)\|('a'...'z'-\_)| | The remaining three encoded characters. Any alphabetic characters must be consistently upper- or lower-case (to distinguish customer- vs. service-managed secrets).
+|encodedKey[^12]|'A'...'_'|Represents the year of key allocation, 'A' (2024) to '_' (2087)|
+|encodedKey[^13]|'A'...'L'|Represents the month of key allocation, 'A' (January) to 'L' (December)|
+|encodedKey[^14]|'A'...'Z'\|'a'..'e'|Represents the day of key allocation, 'A' (0) to 'e' (31)|
+|encodedKey[^15]|'A'...'X'|Represents the hour of key allocation, 'A' (hour 0 or midnight) to 'X' (hour 23). [TBD: This value could be used for half hour increments instead].
```
nguerrera, line 83: This is still missing things after timestamp. Also, the indices look off `^15..^12` is both part of the provider signature and the timestamp.

michaelcfanning:
Shouldn't the pseudocode for generating the keys/hmacs be in `docs` by this definition? this information relates directly to our standard, not anything specific to working in this repo.

---
In reply to: [2590905374](https://github.com/microsoft/cask/pull/26#issuecomment-2590905374) [](http://example.com/codeflow?ancestors=2590905374)

PR #27, opened 2025-01-16
Update implementation to latest documented format

nguerrera:
* Add raw C3ID of primary key to hashes.
* Add all four bytes of CRC32.
* Add CASK version number.
* Add CASK kind (key or hash + size).
* Add day and hour to timestamp.
* Use current time in hash timestamp, and make 'CompareHash' handle that.
* Remove 'allocatorCode' and 'secretEntropyInBytes'.
* Move magic numbers to constants/properties.
* Remove 'ThreeByteSequence' from tests. We don't actually need this helper with the way the format finally landed.

This does not include HMACSHA384 support that is implied by the latest docs. That will be added separately and will likely involve some API refactoring.


docs/PrimaryKey.md
```diff
@@ -57,14 +57,14 @@
...
 |decodedKey[33..^15]|0...255|0x0...0xFF|00000000b...11111111b|Provider-defined data, comprising 0 or more 3-byte sequences, of arbitrary interpretation.
 |decodedKey[^15..^12]| 37, 4, 9  |0x25, 0x04, 0x09| 00100101b, 00000100b, 00001001b | Decoded 'JQQJ' signature.
 |decodedKey[^12..^9]|0...255|0x0...0xFF|00000000b...11111111b| Provider identifier, e.g. , '0x4c', '0x44', '0x93' (base64 encoded as 'TEST')
 |decodedKey[^9..^6]||||Time stamp data encoded in 4 six-bit blocks for YMDH.
-|decodedKey[^6] >> 2|0, 28, 32|0x00, 0x1c, 0x20|00000000b...11111100b| Leading 6 bits comprises kind enum followed by 2 bits of reserved padding. key[key.Length - 6] & 0xfc == 0.
-|decodedKey[^5] >> 4]|0|0xFF|00000000b...11110000b| Leading 4 bits comprise 4 bits of reserved version information + 4 bits of zero padding (to preserve consistent rendering of the subsequence checksum data).
+|decodedKey[^6]|0, 28, 32|0x00, 0x1c, 0x20|00000000b, 00011100b, 00100000b| Leading 6 bits comprises kind enum followed by 2 bits of reserved zero padding.
+|decodedKey[^5]|0|0x00|00000000b| Leading 4 bits comprise version number (currenty 0 = Cask 1.0), followed by 4 bits of reserved zero padding.
```
michaelcfanning, line 67: >currenty [](http://example.com/codeflow?start=74&length=8)

s/be currently

michaelcfanning, line 67: >Leading [](http://example.com/codeflow?start=34&length=7)

So! if we say that the two bits of padding preceding this aren't part of the version, then that full 6 bit character isn't interpretable on its own. Just saying.

src/Cask/Cask.cs
```diff
@@ -69,33 +67,30 @@ public static bool IsCask(ReadOnlySpan<char> key)
     /// <summary>
     /// Validates that the provided UTF8-encoded byte sequence represents a valid Cask key.
     /// </summary>
-    public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUTF8)
+    public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUtf8)
     {
-        // NOTE: Since all valid Cask keys are ASCII-safe, we use the we can
-        //       safely apply `char` limits as `byte` limits.
-        if (keyUTF8.Length < MinKeyLengthInChars || keyUTF8.Length > MaxKeyLengthInChars || !Is4CharAligned(keyUTF8.Length))
+        if (keyUtf8.Length < MinKeyLengthInChars || keyUtf8.Length > MaxKeyLengthInChars || !Is4CharAligned(keyUtf8.Length))
```
michaelcfanning, line 72: >keyUtf8 [](http://example.com/codeflow?start=12&length=7)

Oh! so you're going to depart from .NET's own casing? haha. remember the guidelines state that if a term like `COM` port is an industry norm, not adhering to strict casing guidelines is ok.

I shouldn't troll like this. :) Change is fine.

src/Cask/Cask.cs
```diff
@@ -119,193 +114,193 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
             return false;
         }
 
-        // Check signature: [0x25, 0x04, 0x09] is JQQJ decoded from base64 to bytes
-        ReadOnlySpan<byte> signature = keyBytes[^12..^8];
-        if (0x25 != signature[0] || 0x04 != signature[1] || 0x09 != signature[2])
+        // Check CASK signature
```
michaelcfanning, line 117: >signature [](http://example.com/codeflow?start=22&length=9)

I like periods at the end of sentences. I think you should recap the signature, why not, it's `JQQJ` in the comment. It's nice to co-locate simple information like this where the literal value is elided due to use of a dedicated constant.

IMO

src/Cask/CaskVersion.cs
```diff
@@ -2,12 +2,11 @@
...
+
+public enum CaskVersion
 {
     /// <summary>
     /// Specifies version 1.0.0 of CASK.
     /// </summary>
-    OneZeroZero,
-
-    Reserved = 0xfe,
+    OneZeroZero = 0,
```
michaelcfanning, line 11: >OneZeroZero [](http://example.com/codeflow?start=4&length=11)

We have broken a general rule here, to insist that expressed data is non-zero to distinguish the uninitialized case.  For a v1 though I tend to think this is just fine.

This assumes that any changes requiring a version change would materially alter a key such that it would recognizably no longer be a v0 key if someone neglected to update the version bit.

Also, if we have relevant API that requires specifying a version this is more proof against this class of problem (as opposed to default/optional args that easily get set to zero).

Just talking aloud.

src/Cask/Helpers.cs
```diff
@@ -49,12 +43,74 @@ public static bool Is4CharAligned(int charLength)
...
+        Debug.Assert(Is3ByteAligned(providerDataLengthInBytes));
+        Debug.Assert(Is3ByteAligned(hashLengthInBytes));
+        return hashLengthInBytes;
+    }
+
+    public static KeyKind CharToKind(char kindChar)
     {
-        Debug.Assert(Is3ByteAligned(secretEntropyInBytes), "secretEntropyInBytes should have been rounded up to 3-byte alignment already.");
-        Debug.Assert(Is3ByteAligned(providerDataLengthInBytes), "providerDataLengthInBytes should have been validated to 3-byte aligned already.");
+        return (KeyKind)(kindChar - 'A');
```
michaelcfanning, line 66: >KeyKind [](http://example.com/codeflow?start=16&length=7)

I'm not saying we should do it but I note we could harden this code to attempt to parse the incoming data so that we'd throw on undefined vals.

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// <summary>
+    /// The bytes that make up the CASK signature. "JQQJ", base64-decoded.
+    /// </summary>
+    public static ReadOnlySpan<byte> CaskSignatureBytes => [0x25, 0x04, 0x09];
+
+    /// <summary>
+    /// Current version of CASK spec that we implement.
+    /// </summary>
+    public const int CaskVersion = 0;
```
michaelcfanning, line 25: >CaskVersion [](http://example.com/codeflow?start=21&length=11)

Do we need this int now that we have an enum?

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// </summary>
+    public const int CaskSignatureSizeInBytes = 3;
+
+    /// <summary>
+    /// The number of bytes used to store the timestamp in a key.
+    /// </summary>
+    public const int TimestampSizeInBytes = 3;
+
+    /// <summary>
+    /// The number of bytes  the provider signature.
```
michaelcfanning, line 38: >   [](http://example.com/codeflow?start=27&length=2)

hey! there are two spaces between `bytes  the`

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// </summary>
+    public const int ProviderSignatureSizeInBytes = 3;
+
+    /// <summary>
+    /// The nubmer of bytes reserved in the key footer for future use.
```
michaelcfanning, line 43: >nubmer [](http://example.com/codeflow?start=12&length=6)

whatzanubmer?

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// </summary>
+    public const int FixedKeyComponentSizeInBytes =
+        CaskSignatureSizeInBytes +
+        ProviderSignatureSizeInBytes +
+        TimestampSizeInBytes +
+        VersionAndKindSizeInBytes +
+        Crc32SizeInBytes;
+
+    /// <summary>
+    /// The number of bytes in the fixed componet of a hash key, from the C3ID to the end of the key.
```
michaelcfanning, line 63: >componet [](http://example.com/codeflow?start=41&length=8)

`component`

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// </summary>
+    public const int FixedHashComponentSizeInBytes = FixedKeyComponentSizeInBytes + CrossCompanyCorrelatingId.RawSizeInBytes;
+
+    /// <summary>
+    /// The number of bytes of entropy in a primary key.
+    /// Currently fixed, but may become configurable in future versions.
```
michaelcfanning, line 69: Maybe append:

`32-bytes (256 bits) of entropy generated by a cryptographically secure RNG are currently deemed unbreakable even in a post-quantum world.`

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// </summary>
+    public const int SecretEntropyInBytes = 32;
+
+    /// <summary>
+    /// The size of the entropy in a primary after padding to 3-byte alignment.
+    /// </summary>
+    public const int PaddedSecretEntropyInBytes = 33;
```
michaelcfanning, line 76: >33 [](http://example.com/codeflow?start=50&length=2)

The constants as they are nice and readable. Alternately:

`RoundUpTo3ByteAlignment(SecretEntropyInBytes)`

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// <summary>
+    /// The maximum amount of bytes that the implementation will stackalloc.
+    /// </summary>
+    public const int MaxStackAlloc = 256;
+
+    /// <summary>
+    /// The range of byte indices in a key for the bytes that contain the CASK signature.
+    /// </summary>
+    public static Range CaskSignatureByteRange => ^15..^12;
```
michaelcfanning, line 91: >Range [](http://example.com/codeflow?start=18&length=5)

um. sweet.

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// </summary>
+    public static Range ProviderSignatureByteRange => ^12..^9;
+
+    /// <summary>
+    /// The range of byte indices in a key for the bytes that contain the timestamp.
+    /// </summary>
+    public static Range TimestampByteRange => ^9..^6;
+
+    /// <summary>
+    /// The index of the byte in a key that contains the key kind.
```
michaelcfanning, line 104: >key kind [](http://example.com/codeflow?start=57&length=8)

making the same sorry point here that this byte also contains 2 bits of the encoded char where the version is expressed.

we can abandon these 2 bits for the sake of simplicity for now and probably be ok. the version would have to increase past those four bits before we panicked and wanted to recover them.

if in future we decided to put that range to a different purpose, then each encoded character would have to be able to express the possible combinations of values. and i guess that's fine, after all, no matter what, base64 can cover all 6 bits exhaustively.

so at this point, we've apparently reserved two trailing bits of padding in one byte and then announced the leading four bits in the next comprise the version.

isn't just simpler for v1 to say, for version 1 all these bytes are reserved and must be zeroed out. that's it.

in the future, when one of those bits gets set to zero, we can define their use further. why speak ahead of ourselves?

this comment wins the award for most words on the least significant detail.

nguerrera: > isn't just simpler for v1 to say, for version 1 all these bytes are reserved and must be zeroed out. that's it.

Yes, and isn't that what we've done?

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+    public static Index VersionByteIndex => ^5;
+    /// <summary>
+    /// The range of byte indices in a key for the bytes that contain the CRC32 of the key.
+    /// </summary>
+    public static Range Crc32ByteRange => ^4..;
+
+    /// <summary>
+    /// The range of byte indices in a hash for the bytes that contain the C3ID of the secret.
+    /// </summary>
+    public static Range C3IdByteRange => ^27..^15;
```
michaelcfanning, line 120: >C3IdByteRange [](http://example.com/codeflow?start=24&length=13)

it's up to you, but i'd tend to put this data earlier, immediately preceding `CaskSignatureRange`. The reason is that that's where it is in the key.

we should do the same in BNF/docs, put this data where it will be for an HMAC and note that it's optional.

btw - seeing these ranges and how (relatively!) clear things are to decompose the encoded and bytewise keys. i like it.

src/Cask/InternalConstants.cs
nguerrera: We do not. I had it in this branch before I got your enum.

src/Cask/InternalConstants.cs
michaelcfanning, line 120: >C3IdByteRange [](http://example.com/codeflow?start=24&length=13)

you don't have a range defined for the key optional data but it's computable. as long as have our 33 byte random sequence. that range is identical to the HMAC256 (though a second definition may be warranted as you've done elsewhere).

src/Cask/InternalConstants.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// <summary>
+    /// A bit mask to obtain the reserved bits from the key kind.
+    /// </summary>
+    public const int KindReservedMask = (1 << KindReservedBits) - 1;
+
+    /// <summary>
+    /// The range of chars in a base64-encoded key that hold the Cask signature.
+    /// </summary>
+    public static Range CaskSignatureCharRange = ^20..^16;
```
michaelcfanning, line 145: >CaskSignatureCharRange [](http://example.com/codeflow?start=24&length=22)

you are missing offsets for the encoded flavor.

hey! you offered this review note when I was working on the BNF, just returning the favor.

src/Cask/KeyKind.cs
```diff
@@ -3,49 +3,24 @@
...
-
+    Key256Bit = ('A' - 'A'), // Base64: index 0 == 'A'
+                             // Bytewise: << 2 == 0, 0x00, 0b0000_0000
     /// <summary>
     /// Specifies a CASK hashed signature that incorporates a Message
     /// Authentication code of a derivation input generated by SHA-256.
     /// </summary>
-    Hash256Bit = ('H' - 'A') << 2, // Base64 index 7 == 'H', << 2 == 28, 0x1c, 0b000 11100
+    Hash256Bit = ('H' - 'A'), // Base64 index 7 == 'H'
+                              // Bytewise: << 2 == 28, 0x1c, 0b0001_1100
```
michaelcfanning, line 18: >0b0001_1100 [](http://example.com/codeflow?start=61&length=11)

yes! this is a good convention.

but now that I look at it again, the *real* interest of this comment is to communicate our reserved zeros. i.e., the 4 bit separation is a bit confusing. 

I think I'd just remove the underscore. my care factor -> approaches nil

src/Cask/KeyKind.cs
```diff
@@ -3,49 +3,24 @@
 
 namespace CommonAnnotatedSecurityKeys;
 
-internal enum BytewiseKeyKind : byte
+public enum KeyKind
```
michaelcfanning, line 6: >KeyKind [](http://example.com/codeflow?start=12&length=7)

ok, so you don't want a value you can just slam into a byte?

it's ok to do that if we reserve all the zero bits.

src/Cask/CaskKey.cs
```diff
@@ -116,13 +154,38 @@ public static CaskKey Encode(ReadOnlySpan<byte> bytes)
...
+        if (!IsInitialized)
+        {
+            ThrowOperationOnUnitializedInstance();
+        }
     }
 
     // language=regex
-    private const string RegexPattern = """(^|[^A-Za-z0-9+/-_])([A-Za-z0-9-_]{4}){6,}JQQJ[A-Za-z0-9-_]{12}($|[^A-Za-z0-9+/-_])""";
+    private const string RegexPattern = """(^|[^A-Za-z0-9+/-_])([A-Za-z0-9-_]{4}){6,}JQQJ[A-Za-z0-9-_]{16}($|[^A-Za-z0-9+/-_])""";
     private const RegexOptions RegexFlags = RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.CultureInvariant;
```
michaelcfanning, line 189: >RegexFlags [](http://example.com/codeflow?start=31&length=10)

We should add `NonBacktracking ` for builds that allow it.

src/Tests/Cask.Benchmarks/BenchmarkTestData.cs
```diff
@@ -9,7 +9,6 @@ internal static class BenchmarkTestData
 {
     public const string TestDerivationInput = "Lorem ipsum dolor sit amet, consectetur adipiscing elit";
     public const string TestProviderSignature = "TEST";
-    public const string TestAllocatorCode = "88";
     public const string TestProviderData = "0123456789ABCDEF";
     public const int TestSecretEntropyInBytes = 32;
```
michaelcfanning, line 13: >32 [](http://example.com/codeflow?start=48&length=2)

you are internals visible to averse and hence didn't use `InternalConstants.SecretEntropyInBytes`  here. ok.

src/Tests/Cask.Tests/ThreeByteSequenceTests.cs
```diff
@@ -1,54 +0,0 @@
...
-using Xunit;
-
-namespace CommonAnnotatedSecurityKeys.Tests;
-
-public class ThreeByteSequenceTests
-{
-    private const string Base64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
-
-    [Fact]
-    public void ThreeByteSequence_ExhaustiveBase64Url()
```
michaelcfanning, line 15: >ThreeByteSequence_ExhaustiveBase64Url [](http://example.com/codeflow?start=16&length=37)

good bye 32 seconds of test execution on .net framework!

src/Cask/InternalConstants.cs
nguerrera: I misunderstood. Yes, it's simpler to say it's all zero now. I will do that.

src/Cask/Cask.cs
nguerrera: Yeah, .NET itself is inconsistent now. Base64Url that is new in the BCL and that we use extensively uses Utf8. I took that as signal that that's the way to go these days, but I had just one place where old habits remained so I changed this one to match.

src/Cask/Helpers.cs
nguerrera: I'll put an assert. We only use this on validated keys.

src/Cask/InternalConstants.cs
nguerrera: I've limited the ranges to the ones the implementation needs for now since this is all internal.

src/Cask/InternalConstants.cs
nguerrera: I'm not, there is `CaskSignatureByteRange` above.

src/Cask/KeyKind.cs
nguerrera: Since your care factor approaches nil, I'm going to leave this. :) My tired eyes have a lot of trouble with binary that isn't broken down into nybbles.

src/Cask/KeyKind.cs
nguerrera: My care factor isn't high, but the base64 index is more like a traditional enum with the only gap being for room for larger keys before the hashes start. I'll keep this.

src/Cask/CaskKey.cs
nguerrera: Interestingly, GeneratedRegexAttribute doesn't really support NonBacktracking.

The generated code just compiles the regex at runtime, same as our .NET Framework polyfill when we add that. 

Resolving since this is pre-existing and unrelated to this big change, but if we need to use NonBacktracking then we should probably remove the complexity of the GeneratedRegexAttribute + polyfill.

https://github.com/dotnet/runtime/issues/61903

nguerrera:
@michaelcfanning All feedback addressed. Please take another look.

src/Cask/Cask.cs
```diff
@@ -113,199 +110,200 @@ public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUTF8)
...
-        // Use another span like a pointer, moving it forward as we write data.
-        Span<byte> destination = keyBytes;
+        Span<byte> key = stackalloc byte[keyLengthInBytes];
 
         // Entropy
-        FillRandom(destination[..secretEntropyInBytes]);
-        destination = destination[secretEntropyInBytes..];
+        FillRandom(key[..SecretEntropyInBytes]);
+
+        // Padding
```
michaelcfanning, line 164: >Padding [](http://example.com/codeflow?start=11&length=7)

I wish I could get you to obsessively add periods at the end of all comments. I'll talk to Copilot about it, I can probably outsource this kind of NIT nagging to the machine world.

src/Cask/Cask.cs
```diff
@@ -113,199 +110,200 @@ public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUTF8)
...
-        Crc32.Hash(keyBytes[..^3], crc32);
-        crc32[..3].CopyTo(checksumDestination);
+        if (timestamp.IsEmpty)
+        {
+            DateTimeOffset now = GetUtcNow();
+            ValidateTimestamp(now);
+            ReadOnlySpan<char> chars = [
+                Base64UrlChars[now.Year - 2024], // years since 2024
+                Base64UrlChars[now.Month - 1],   // zero-indexed month
+                Base64UrlChars[now.Day],         // zero-indexed day
```
michaelcfanning, line 251: >zero-indexed day [](http://example.com/codeflow?start=52&length=16)

are you sure you zero-indexed this one?

nguerrera: Good catch.

src/Cask/Cask.cs
```diff
@@ -113,199 +110,200 @@ public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUTF8)
...
-
-        // Decode candidate hash
-        int candidateHashLengthInBytes = Base64CharsToBytes(candidateHash.ToString().Length);
-        if (candidateHashLengthInBytes != length)
+        ThrowIfUnitialized(candidateHash);
+        ThrowIfNotHash(candidateHash);
+        ThrowIfUnitialized(secret);
+        ThrowIfNotPrimary(secret);
+
+        // Check if sizes match
```
michaelcfanning, line 290: >match [](http://example.com/codeflow?start=26&length=5)

Periods. Please. Haha.

src/Cask/Cask.cs
nguerrera: How about this: please use the suggested change feature so I can just click accept on these. :)

src/Cask/CaskKey.cs
```diff
@@ -11,18 +12,46 @@ namespace CommonAnnotatedSecurityKeys;
...
+    /// Indicates if the key is initialized, and not the default struct value.
+    /// </summary>
+    [MemberNotNullWhen(true, nameof(_key))]
+    public bool IsInitialized => _key != null;
+
+    public KeyKind Kind
+    {
+        get
+        {
+            ThrowIfUnitialized();
```
michaelcfanning, line 34: >ThrowIfUnitialized [](http://example.com/codeflow?start=12&length=18)

s/b `Uninitialized`

src/Cask/CaskKey.cs
```diff
@@ -116,13 +145,38 @@ public static CaskKey Encode(ReadOnlySpan<byte> bytes)
...
+
+        int bytesWritten = Base64Url.DecodeFromChars(_key.AsSpan(), destination);
+        Debug.Assert(bytesWritten == SizeInBytes);
+    }
+
     public override string ToString()
     {
-        return _key ?? "";
+        // Throwing violates the contract for ToString(), but it's a safety
+        // measure to block a bug from causing an unitialized key from being
```
michaelcfanning, line 160: >unitialized [](http://example.com/codeflow?start=50&length=11)

misspelled

nguerrera: This gives me empathy for everyone that thinks I'm Nick Guerra.

src/Cask/CaskKey.cs
michaelcfanning: The NonBacktracking option opts into a regex that provides significant perf gains, which s/be measurable.

src/Cask/CaskKey.cs
```diff
@@ -116,13 +145,38 @@ public static CaskKey Encode(ReadOnlySpan<byte> bytes)
...
+        // used as a production key.
+        //
+        // This is not considered final and needs more thought. We may move
+        // access to the underlying string elsewhere.
```
michaelcfanning, line 164: >elsewhere [](http://example.com/codeflow?start=43&length=9)

Hm. Interesting topic. Can we better control/get ahead of this via explicit/implicit type operators?

For debug builds, it'd be nice for CaskKey.ToString() to provide a breakdown of the key and its components. For release, ToString could dip into a managed underlying string if we take that route.

src/Cask/CaskKey.cs
```diff
@@ -116,13 +145,38 @@ public static CaskKey Encode(ReadOnlySpan<byte> bytes)
...
+    {
+        if (!IsInitialized)
+        {
+            ThrowOperationOnUnitializedInstance();
+        }
     }
 
     // language=regex
-    private const string RegexPattern = """(^|[^A-Za-z0-9+/-_])([A-Za-z0-9-_]{4}){6,}JQQJ[A-Za-z0-9-_]{12}($|[^A-Za-z0-9+/-_])""";
+    private const string RegexPattern = """(^|[^A-Za-z0-9+/-_])([A-Za-z0-9-_]{4}){6,}JQQJ[A-Za-z0-9-_]{16}($|[^A-Za-z0-9+/-_])""";
```
michaelcfanning, line 179: >RegexPattern [](http://example.com/codeflow?start=25&length=12)

this regex pattern troubles me because we don't consistently escape the standalone hyphen char. let's do that.

src/Cask/Helpers.cs
michaelcfanning: I keep forgetting this is a public method on an internal class.

src/Cask/Cask.cs
michaelcfanning: gotcha,

src/Cask/Cask.cs
michaelcfanning: This kind of bug is why I tend to author the (occasionally pilloried) tests that I do which exhaustively produce and validate all possible ranges for data of this kind.

src/Cask/CaskKey.cs
michaelcfanning: sometimes shorted to Nick Grr

PR #28, opened 2025-01-23
Update pseudo-code to latest design

nguerrera:
Also move the pseudo-code from `devdocs` to `docs`.


docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
...
+    - N bytes for provider data
+    - 12 bytes for C3ID
+    - 3 bytes for CASK signature
+    - 3 bytes for provider signature
+    - 3 bytes for timestamp 
+    - 1 reserved byte
+    - 1 byte for size and kind
+    - 4 bytes for CRC32 checksum
+1. Compute HMACSHA256 or HMACSHA384 using the CASK secret as the key and the provided derivation input as the message. Store the result at the beginning of the CASK hash.
+1. If using HMACSHA256, write zero to the next byte (padding to maintain 3-byte alignment).
```
nguerrera, line 16:
```suggestion
1. If using HMACSHA256, write 0x00 to the next byte (padding to maintain 3-byte alignment).
```

docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
...
+1. Compute C3ID of the CASK secret. Store the result in the next 12 bytes.
+1. Write the CASK signature [0x25, 0x04, 0x09] ("JQQJ", base64-decoded) to the next 3 bytes.
+1. Copy the provider signature from the CASK secret to the next 3 bytes.
+1. Obtain the current date and time in UTC. Store it in 4 characters, YMDH:
+    - Y = base64url encoding of (Year - 2024).
+    - M = base64url encoding of zero-based month.
+    - D = base64url encoding of zero-based hour.
+    - H = base64url encoding of zero-based day.
+1. base64url-decode YMDH and store the result in the next 3 bytes.
+1. Write zero to next byte (reserved).
```
nguerrera, line 27:
```suggestion
1. Write 0x00 to the next byte (reserved).
```

docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
...
+1. Write 0x1C to indicate HMACSHA256 or 0x20 to indicate HMACSHA384 to the next byte.
+1. Compute the CRC32 of all CASK hash bytes written above (everything but last 4 bytes.) Store the result in little-endian byte order in the last 4 bytes.
```
nguerrera, line 29:
```suggestion
1. Compute the CRC32 of all CASK hash bytes written above (everything but last 4 bytes). Store the result in little-endian byte order in the last 4 bytes.
```

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
...
+1. base64url decode provider data and store the result in the next N bytes.
+1. Write CASK signature [0x25, 0x04, 0x09] ("JQQJ", base64-decoded) to the next 3 bytes.
+1. base64url decode provider signature and store the result in the next 3 bytes.
+1. Obtain the current date and time in UTC. Store it in 4 characters, YMDH:
+    - Y = base64url encoding of (Year - 2024).
+    - M = base64url encoding of zero-based month.
+    - D = base64url encoding of zero-based hour.
+    - H = base64url encoding of zero-based day.
+1. base64url-decode YMDH and store the result in the next 3 bytes.
+1. Write zero to the next byte (reserved).
```
nguerrera, line 30:
```suggestion
1. Write 0x00 to the next byte (reserved).
```

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
...
+    - 1 padding byte
+    - N bytes for provider data
+    - 3 bytes for CASK signature
+    - 3 bytes for provider signature
+    - 3 bytes for timestamp
+    - 1 reserved byte
+    - 1 byte for size and kind
+    - 4 bytes for CRC32 checksum
+1. Generate 256 bits of cryptographically secure random data. Store the result at the beginning of the generated key.
+1. Write zero to the next byte (padding to maintain 3-byte alignment).
```
nguerrera, line 20:
```suggestion
1. Write 0x00 to the next byte (padding to maintain 3-byte alignment).
```

docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
+# GenerateHash Pseudo-code
+
+1. Examine the provided secret to determine if it is a CASK primary key. If it is not, return an error.
+1. Let N = number of bytes in the provider data of the CASK secret. (N may be 0 if there is no provider data.)
```
michaelcfanning, line 4: >bytes [](http://example.com/codeflow?start=21&length=5)

`number of 3-byte segments of provider data` or some such. the bytes here must 3 byte align, IOW

michaelcfanning: ah! this is already accomplished because the primary key is valid. got it.

docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
...
+1. Allocate storage for the CASK hash:
+    - 33 bytes for padded HMACSHA256 or 48 bytes for HMACSHA384
+    - N bytes for provider data
```
michaelcfanning, line 7: >data [](http://example.com/codeflow?start=27&length=4)

maybe add a comment here (because the CASK primary key has been determined to be valid, 'N' will comprise a number that is divisible by 3

docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
+# GenerateHash Pseudo-code
+
+1. Examine the provided secret to determine if it is a CASK primary key. If it is not, return an error.
```
nguerrera, line 3:
```suggestion
1. Examine the provided secret to determine if it is a valid CASK primary key. If it is not, return an error.
```

docs/GenerateHashPseudoCode.md
nguerrera: I added the word valid now :)

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
+# GenerateKey Pseudo-Code
+
+1. Validate input. Return an error if any of the following are NOT true:
+    - Provider signature is exactly 4 characters long.
+    - Provider signature consists entirely of characters that are valid in base64url encoding.
+    - Provider data (if any) has a length that is a multiple of 3 characters and no more than 32 characters.
```
nguerrera, line 6: I think I go this wrong. Fixing

nguerrera, line 6:
```suggestion
    - Provider data (if any) has a length that is a multiple of 4 characters and no more than 32 characters.
```

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
...
+    - Provider data (if any) consists entirely of characters that are valid in base64url encoding.
+1. Let N = the length of the base64url-decoded provider data. Number of characters in the provider data, divided by 3, times 4.
```
nguerrera, line 8:
```suggestion
1. Let N = the length of the base64url-decoded provider data. (Number of characters in provider data, divided by 4, times 3.
```

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
+# GenerateKey Pseudo-Code
+
+1. Validate input. Return an error if any of the following are NOT true:
+    - Provider signature is exactly 4 characters long.
+    - Provider signature consists entirely of characters that are valid in base64url encoding.
+    - Provider data (if any) has a length that is a multiple of 4 characters and no more than 32 characters.
+    - Provider data (if any) consists entirely of characters that are valid in base64url encoding.
+1. Let N = the length of the base64url-decoded provider data. (Number of characters in provider data, divided by 4, times 3.
```
nguerrera, line 8:
```suggestion
1. Let N = the length of the base64url-decoded provider data. (Number of characters in provider data, divided by 4, times 3.)
```

docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
...
+    - 3 bytes for timestamp 
+    - 1 reserved byte
+    - 1 byte for size and kind
+    - 4 bytes for CRC32 checksum
+1. Compute HMACSHA256 or HMACSHA384 using the CASK secret as the key and the provided derivation input as the message. Store the result at the beginning of the CASK hash.
+1. If using HMACSHA256, write 0x00 to the next byte (padding to maintain 3-byte alignment).
+1. Copy provider data from the CASK secret to the next N bytes.
+1. Compute C3ID of the CASK secret. Store the result in the next 12 bytes.
+1. Write the CASK signature [0x25, 0x04, 0x09] ("JQQJ", base64-decoded) to the next 3 bytes.
+1. Copy the provider signature from the CASK secret to the next 3 bytes.
```
michaelcfanning, line 20: >signature [](http://example.com/codeflow?start=21&length=9)

NIT: we agreed to call this provider id since it will consist of multiple components, a provider piece (like `AZ` for 'Azure') and then a specific allocator id (like `ST` for 'Storage')

docs/GenerateHashPseudoCode.md
```diff
@@ -0,0 +1,33 @@
...
+1. Obtain the current date and time in UTC. Store it in 4 characters, YMDH:
+    - Y = base64url encoding of (Year - 2024).
+    - M = base64url encoding of zero-based month.
+    - D = base64url encoding of zero-based hour.
+    - H = base64url encoding of zero-based day.
+1. base64url-decode YMDH and store the result in the next 3 bytes.
+1. Write 0x00 to the next byte (reserved).
+1. Write 0x1C to indicate HMACSHA256 or 0x20 to indicate HMACSHA384 to the next byte.
+1. Compute the CRC32 of all CASK hash bytes written above (everything but last 4 bytes). Store the result in little-endian byte order in the last 4 bytes.
```
michaelcfanning, line 29: do we have endianness tests yet, btw? just curious.

docs/GenerateKeyPseudoCode.md
michaelcfanning: I believe it is 20 characters. If we allow 20 characters, this is 15 bytes of optional data. this will take the key size to 63 bytes. we can't exceed this (i.e., flip over to 66 bytes) as 66 bytes will exceed the 512 bits maximum for the HMAC secret.

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
+# GenerateKey Pseudo-Code
+
+1. Validate input. Return an error if any of the following are NOT true:
+    - Provider signature is exactly 4 characters long.
```
michaelcfanning, line 4: >signature [](http://example.com/codeflow?start=15&length=9)

if we preserve HIS v2 conventions, the leading character of every provider id *must* be an alpha character. Furthermore, this alpha character establishes a casing for the provider id that must be consistent with any other alpha chars in the signature. A signature with a leading upper-case character comprises a user-managed key (and all other chars in the sig must be upper-case). A signature with a leading lower-case char comprises a service-managed key (and all other chars in the sig must also be lowercase).

docs/GenerateKeyPseudoCode.md
michaelcfanning: >characters [](http://example.com/codeflow?start=66&length=10)

are we working in bytes or characters for this description, I'm getting a little confused.

if the provider data is chars, the length is a multiple of 4, not more than 20 encoded chars. if it is bytes, the length is a multiple of 3, no more than 15 bytes.

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
...
+    - Provider signature consists entirely of characters that are valid in base64url encoding.
+    - Provider data (if any) has a length that is a multiple of 3 characters and no more than 32 characters.
+    - Provider data (if any) consists entirely of characters that are valid in base64url encoding.
```
michaelcfanning, line 7: >characters that are valid [](http://example.com/codeflow?start=50&length=25)

ok, this seems to indicate we are working with chars.

docs/GenerateKeyPseudoCode.md
```diff
@@ -0,0 +1,36 @@
...
+    - 3 bytes for timestamp
+    - 1 reserved byte
+    - 1 byte for size and kind
+    - 4 bytes for CRC32 checksum
+1. Generate 256 bits of cryptographically secure random data. Store the result at the beginning of the generated key.
+1. Write 0x00 to the next byte (padding to maintain 3-byte alignment).
+1. base64url decode provider data and store the result in the next N bytes.
+1. Write CASK signature [0x25, 0x04, 0x09] ("JQQJ", base64-decoded) to the next 3 bytes.
+1. base64url decode provider signature and store the result in the next 3 bytes.
+1. Obtain the current date and time in UTC. Store it in 4 characters, YMDH:
```
michaelcfanning, line 24: >UTC [](http://example.com/codeflow?start=39&length=3)

should we be clear that the UTC should be obtained and cached to produce this data? to help folks avoid bugs where Utc.Now flickers across a boundary of some kind.

1. Obtain and cache the current date and time in UTC.

docs/GenerateHashPseudoCode.md
nguerrera: We do not.

docs/GenerateKeyPseudoCode.md
nguerrera: I had a bug here with 3 and 4 inverted, but the limits in the code are currently, 32 chars / 24 bytes.

docs/GenerateKeyPseudoCode.md
nguerrera: OK, should we preserve these? I found the upper and lower case thing confusing when I touched something related for the first time in HISv2, FWIW.

The current description matches the current code. And the current code was chosen by me when I discovered that the initial draft did no validation.

If we want to change this, I think it would be best to do it in its own PR that updates the code, tests, and pseudo-code.

docs/GenerateKeyPseudoCode.md
michaelcfanning: I don't believe this is correct. Validate my thinking.

33 random + 6 signatures + 3 timestamp + 6 (reserved + checksum)
That's 48 bytes. Now, we have a natural limit on this key of 64 bytes, otherwise the key risks triggering HMAC hashing on use as an initializer. Because of our 3 byte alignment, our natural limit is 63 bytes. 63 - 48 bytes == 15 byte limit on optional data.

docs/GenerateHashPseudoCode.md
nguerrera: OK, it's called provider signature all over the code still. I will send a separate PR to rename it to id.

docs/GenerateKeyPseudoCode.md
michaelcfanning: Sure, makes sense. Whatever is best to clarify we obtain a single timestamp and perform the conversion. I like your language.

nguerrera: Done.

docs/GenerateKeyPseudoCode.md
nguerrera: I am avoiding changing the code in this change. I believe what I said is correct with respect to the current implementation. I will change the limit in code and pseudo-code in a follow-up.

docs/GenerateHashPseudoCode.md
michaelcfanning: I don't know how much I care about this but it'd be good to be consistent.

docs/GenerateKeyPseudoCode.md
michaelcfanning: OK. So, this convention is important/useful for providers but it doesn't have to be a requirement in the standard.

The core problem is ownership/remediation accountability. When a user-managed key is spilled, we want the user to be informed/to remediate. When a service-managed key is spilled, we obviously want to alert the service provider and the provider will benefit from time/air cover if possible to drive remediation.

Imagine a secret spilled into public OSS. For a user-managed key, you would reasonably inform the repo owner of the exposure. For a server managed secret, you might not want that kind of alerting, instead, you'd want a direct notification sent to the platform only (GitHub sends these notices to service owners today). 

So let's check in what you have, and perhaps we have some supplemental content in our repo for advanced/nuanced extensions to the core requirements. How to encode a key lifetime is another example where some suggested techniques might make sense without having a hard standard.

docs/GenerateKeyPseudoCode.md
michaelcfanning: makes sense, let's get this in and keep iterating.

PR #41, opened 2025-02-25
Provider key kind as char.

### LINKED ISSUE #33: Update `providerKeyKind` `GenerateKey` parameter from `string` to `char`
And also take a good look at the use of optional args in this signature and its interface version.

The provider key kind in theory could be optional. On the other hand, it is very frequently utilized by providers and so there wouldn't be a lot of friction to requiring it (and a non-optional form could arguably be useful for discovery of this feature.

michaelcfanning:
Code change to resolve #33.

Currently, provider key kind is a string, which we check to ensure is non-null, has length of 1, is url safe base64, etc.

The change is to accept a char value only for the provider key kind (which ends up being encoded into 6 bits of the key). This is a more appropriate data type and also simplifies validation and some other logic.

### DIFF
--- a/src/Cask/Helpers.cs
+++ b/src/Cask/Helpers.cs
@@ -73,8 +73,8 @@ internal static class Helpers
     {
         int base64Index;
 
-        const int uppercaseZIndex = 25; // 'Z' - 'A';
-        const int lowercaseZIndex = 51; // 'z' - 'a';
+        const int uppercaseZIndex = 'Z' - 'A';
+        const int lowercaseZIndex = 'z' - 'a';
 
         if (providerKind >= 'A' && providerKind <= 'Z')
         {
--- a/src/Cask/Helpers.cs
+++ b/src/Cask/Helpers.cs
@@ -59,7 +59,7 @@ internal static class Helpers
 
     public static CaskKeyKind CharToKind(char kindChar)
     {
-        Debug.Assert(kindChar == 'P' || kindChar == 'H',
+        Debug.Assert(kindChar == 'D' || kindChar == 'H' || kindChar == 'P',
                      "This is only meant to be called using the kind char of a known valid key.");
         return (CaskKeyKind)(kindChar - 'A');
     }
@@ -130,7 +130,7 @@ internal static class Helpers
         }
 
         kind = (CaskKeyKind)(value >> CaskKindReservedBits);
-        return true;
+        return kind is CaskKeyKind.DerivedKey or CaskKeyKind.HMAC or CaskKeyKind.PrimaryKey;
     }
 
     public static bool IsValidForBase64Url(string value)
--- a/src/Cask/Helpers.cs
+++ b/src/Cask/Helpers.cs
@@ -3,7 +3,6 @@
 
 global using static CommonAnnotatedSecurityKeys.Helpers;
 
-using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
@@ -70,36 +69,32 @@ internal static class Helpers
         return (byte)((int)kind << CaskKindReservedBits);
     }
 
-    public static byte ProviderKindToByte(string providerKind)
+    public static byte ProviderKindToByte(char providerKind)
     {
-        Debug.Assert(providerKind?.Length == 1, "Provider kind should be a single character.");
-
         int base64Index;
 
         const int uppercaseZIndex = 25; // 'Z' - 'A';
         const int lowercaseZIndex = 51; // 'z' - 'a';
 
-
-        char providerKindChar = providerKind[0];
-        if (providerKindChar >= 'A' && providerKindChar <= 'Z')
+        if (providerKind >= 'A' && providerKind <= 'Z')
         {
-            base64Index = providerKindChar - 'A';
+            base64Index = providerKind - 'A';
         }
-        else if (providerKindChar >= 'a' && providerKindChar <= 'z')
+        else if (providerKind >= 'a' && providerKind <= 'z')
         {
-            base64Index = providerKindChar - 'a' + uppercaseZIndex;
+            base64Index = providerKind - 'a' + uppercaseZIndex;
         }
-        else if (providerKindChar >= '0' && providerKindChar <= '9')
+        else if (providerKind >= '0' && providerKind <= '9')
         {
-            base64Index = providerKindChar - '0' + lowercaseZIndex;
+            base64Index = providerKind - '0' + lowercaseZIndex;
         }
-        else if (providerKindChar == '-')
+        else if (providerKind == '-')
         {
             base64Index = 62;
         }
         else
         {
-            Debug.Assert(providerKindChar == '_', "Provider kind should be a valid Base64Url char.");
+            Debug.Assert(providerKind == '_', "Provider kind should be a valid Base64Url char.");
             base64Index = 63;
         }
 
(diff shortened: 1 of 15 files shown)


src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -446,4 +444,25 @@ private void IsCaskVerifyFailure(string key)
             Assert.Throws<FormatException>(() => Cask.IsCaskBytes(keyBytes!));
         }
     }
+
+    private static bool IsValidForBase64Url(char c)
```
michaelcfanning, line 448: >IsValidForBase64Url [](http://example.com/codeflow?start=24&length=19)

This is code I duplicated from the product. In another approach, I could have use internals visible to so that `Helpers` is callable from tests. I did not do so as Nick has expressed a desire to preserve a more stringent separation.

nguerrera: Hmm, my thoughts on IVTA and tests are not quite that crisp in general and have even softened since I expressed this

However, this also happens to be a *perfect* example where it's super-important not to call the product IMHO. If we did, we'd pretty much be using this method to generate the test cases for this method! 

Copy/paste isn't much better. I still need to trust this copy of the product code under test to trust these tests. Good test cases should gain my trust *trivially*.

Here's a small change to what you have that would resolve this concern in this instance:

1. Delete this method.
2. In the loop that calls this do:

```diff
> var validBase64Url = new HashSet<char>("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_");

for (int i = 0; i < 256; i++)
{
< if (IsValidForBase64Url(providerKeyKind)) { continue; }
> if (!validBase64Url.Contains(providerKeyKind)) { continue; }
}
```
I have more to say on this topic, but I'll save it for a forum that isn't on the critical path to accepting a small PR.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -241,9 +237,9 @@ public void CaskSecrets_IsCask_InvalidKey_Whitespace()
         // Replace first 4 characters of secret with whitespace. Whitespace is
         // allowed by `Base64Url` API but is invalid in a Cask key.
         string key = $"    {Cask.GenerateKey("TEST",
-                            "X",
-                            expiryInFiveMinuteIncrements: 12 * 24 * 90, // 90 days.
-                            providerData: null)[4..]}";
+                                             'X',
+                                             expiryInFiveMinuteIncrements: 12 * 24 * 90, // 90 days.
+                                             providerData: null)[4..]}";
```
nguerrera, line 242: I would find this easier to read if it were:

```
string key = Cask.GenerateKey(...);
key = $"    {key[4..]};
```

nguerrera (changes requested):

src/Cask/Helpers.cs
```diff
@@ -70,36 +69,32 @@ public static byte KindToByte(CaskKeyKind kind)
...
 
-    public static byte ProviderKindToByte(string providerKind)
+    public static byte ProviderKindToByte(char providerKind)
     {
-        Debug.Assert(providerKind?.Length == 1, "Provider kind should be a single character.");
-
         int base64Index;
 
         const int uppercaseZIndex = 25; // 'Z' - 'A';
         const int lowercaseZIndex = 51; // 'z' - 'a';
```
nguerrera, line 77:
```suggestion
        const int uppercaseZIndex = 'Z' - 'A';
        const int lowercaseZIndex = 'z' - 'a';
```
From the prior PR, but there's no benefit to hard-coding + comment here. The compiler can do this for you. The generated code will be identical so no runtime cost.

src/Tests/Cask.Tests/CaskSecretsTests.cs
michaelcfanning: We are of a mind, precisely, Product code should be replaced with the simplest, most readable replacement possible, you've generated a very nice one, thanks.

michaelcfanning: Was able to remove this code entirely, I will need to go remind myself why it was required in the first place.

src/Cask/Helpers.cs
michaelcfanning: Thank you compiler.

src/Tests/Cask.Tests/CaskSecretsTests.cs
michaelcfanning: Well, I agree.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -182,6 +180,34 @@ public void CaskSecrets_IsCask_InvalidKey_InvalidSensitiveDataSize()
         IsCaskVerifyFailure(key);
     }
 
+    [Fact]
+    public void CaskSecrets_IsCask_InvalidKey_ProviderKeyKind()
+    {
+        for (char providerKeyKind = char.MinValue; providerKeyKind < char.MaxValue; providerKeyKind++)
+        {
+            if (IsValidForBase64Url(providerKeyKind))
```
nguerrera, line 188: Is this calling the product now and that's why it compiled when you deleted?

src/Tests/Cask.Tests/CSharpCaskTests.cs
```diff
@@ -48,6 +47,20 @@ public bool IsCask(string key)
                 // will throw for this condition.
             }
 
+            if (key.Any((c) => !s_validBase64Url.Contains(c)))
+            {
+                // This condition will only occur if the input passed to
+                // `Base64Url.DecodeFromUtf8` included characters (such as 
+                // whitespace) that are not valid ('printable', i.e.,
+                // non-padding) usr-safe base64 characters. Keys that contain
```
nguerrera, line 55:
```suggestion
                // non-padding) base64url characters. Keys that contain
```

Since the RFC specifically says that this can be referred to as "base64url".

Alternatively, s/usr/url/.

src/Tests/Cask.Tests/CSharpCaskTests.cs
```diff
@@ -68,6 +81,15 @@ public bool IsCask(string key)
                     + string.Join(Environment.NewLine, checks.Select(c => $"  {c.name} -> {c.value}")));
             }
 
+            if (keyBytes == null && result == true)
+            {
+                // See comment above for more information on this condition.
+                Assert.Fail(
+                    "Expected a failure condition as input key bytes did strictly conforming to input key string."
```
nguerrera, line 88:
```suggestion
                    "Expected a failure result as the input string contains invalid base64url characters."
```
Or some other re-wording, I can't decipher this as-is.

src/Tests/Cask.Tests/CSharpCaskTests.cs
```diff
@@ -68,6 +81,15 @@ public bool IsCask(string key)
...
+                    + Environment.NewLine
+                    + $"key: {key}");
+            }
```
nguerrera, line 91: Consider this over `+` and `Environment.NewLine`

``` c#
$"""
Expected a failure condition as input key bytes did strictly conforming to input key string.
key: {key}
"""
```

Note that indentation of closing """ is removed from all lines. I love this feature. This is not important, though. Just a thought in case you hadn't considered it.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -12,10 +10,11 @@
 using static CommonAnnotatedSecurityKeys.Limits;
 
 namespace CommonAnnotatedSecurityKeys.Tests;
-
-[ExcludeFromCodeCoverage]
 public abstract class CaskTestsBase
 {
+    private protected static new HashSet<char> s_validBase64Url =
```
nguerrera, line 15:
```suggestion
    private protected static readonly HashSet<char> s_validBase64Url =
```

Build is failing on the `new` and should be `readonly`.

nguerrera (approved):

Issue #37, opened 2025-02-24
Can we infer sensitive data + optional data size directly from the key length?


michaelcfanning:
If we restrict the amount of optional data that is legal for a CASK secret such that a key + its data can't exceed the minimal size of the next CASK key size, we'd have a simple way of understanding both the size of the sensitive component as well as its optional data. This would simplify validity checks as well as high-performance matching scenarios with look-behind/look-ahead logic after first locating the fixed CASK signature.

Currently, we define 30 bytes of required data:
- CASK signature - 3 bytes
- Provider signature - 3 bytes
- Provider/Cask Kind - 2 bytes
- Correlating id - 16 bytes
- Timestamp + expiry 6 bytes

This leaves us with the following minimal key sizes (as measure by the size of the sensitive component). Currently, the key format requires sufficient space to declare the key size. This happens naturally with key sizes of 16, 32, and 64 bytes but a 48 byte key falls along the base64 encoding 3-byte boundary and therefore requires us to pad that data with an additional 3 bytes for 51 total). If we accept this proposal, we no longer need to pad a 48-byte sensitive component (and so the size below doesn't add the additional 3 bytes).

- 128-bit : 48 bytes (18 bytes sensitives + 30) : 12 bytes of optional data permissible (63 - 48)
- 256-bit : 63 bytes (33 bytes sensitive + 30) :  12 bytes of optional data permissible (78 - 63)
- 384-bit : 78 bytes (48 bytes sensitive + 30) :  15 bytes of optional data permissible (96 - 78)
- 512-bit : 96 bytes (66 bytes sensitive + 30) : 15 bytes (value chosen align with 384 bit keys)

There is a proposal to drop the minutes/expiry in #36 that would expand optional limits to 15 bytes (or 18 in the case of 384-bit or 512-bit secrets).

With the approach above and by limiting the maximal key size to 111 bytes (a 64-byte sensitive component + 30 bytes of required data + 18 bytes optional), a simple set of range checks could be used to validate we have a key, the size of its sensitive component and the size of its optional component.

The change from 12 bytes to 15 bytes is an unfortunate, small bit of complexity/inconsistency that is rooted in our insistence on 3-byte alignments. We mitigate this by permitting the larger key sizes to both have the increased optional data size (rather than dropping the 512-bit key limit to 12 bytes to conform to other key kinds).

Size across the wire for base64-encoded forms:

- 128 bits : minimum of 64 characters to a maximum of 80
- 256 bits : 84 characters to 100
[...]

Issue #36, opened 2025-02-24
Consider removing timestamp + expiry as required data


michaelcfanning:
The extension to the format to include a minutes value followed by an 18-bit expiry (expressed as five minute increments, for a maximal expiry of 910 days) has the following issues:

- Notably, this value encourages runtime interpretation. As a principle, CASK encodes data that may be useful in driving efficient security response and secret management but is not designed to encode information that drive auth flows. Post-exposure, an expiry is arguably useful in that it may provide a clue to a responder whether the key is still exploitable and a window of exploitability. But the nature of the expiry (as a base64-encoded number of five minute increments) is less easy to interpret in this way. We recently removed the checksum data integrity check in CASK and even this mechanism, if it were available, is easily circumvented if an attacker became aware that a 'liveness' check was confirmed directly from a CASK key without first verifying the key itself is well-known.
- The combination of a 6-bit minutes value followed by an 18-bit expiry presents a bit more complexity in encoding/decoding over other 3-byte regions in the format.

To address the above, we could consider removing this data sequence from the standard format and note the design as one possibility for optional data that could be appended to the format.

There is a clear need for short-lived secrets in CASK. It isn't clear that the availability of minutes data adds significant to CASK value beyond the YMDH in the core timestamp. A primary focus for this key kind is real-time detection and redaction in context of logging and core CASK features enable this.

michaelcfanning:
Yes, thank you, this issue is resolved by #43.

PR #43, opened 2025-03-06
Significant revision of cask. Remove expiry. Reorder data. Eliminate CASK kind.

### LINKED ISSUE #35: Confirm or deny the need for a CASK key kind.
Our core format has recently been simplified to eliminate API/etc. for performing cryptographic operations, other than cryptographically secure RNG fill for generating basic keys.

Instead, we allocate a range of a key for sensitive data, that may be useful for cryptographic algorithm output (in addition to RNG filled bytes) such as storing an HMAC or a new key generated by a KDF.

This makes our API simpler and doesn't commit our code to any specific cryptographic technique. Balancing this, we clearly need to describe certain common operations in our spec (for handling HMAC generation and validation, for example) and this need argues to some degree for a convenient, well-tested API. There's no reason we couldn't build some of these vetted helpers, however, outside of the spec - we simply wouldn't encode these techniques into the standard. This allows for a certain resilience in the standard in that the recommended guidance for generating or processing secrets in the future may differ from best practices today.

Generally, secure practices argue against using primary keys, i.e., exclusively preferring derived keys (which encourages a `single key, single use` discipline. And so a specific designation of primary vs. derived could be helpful. The presence of a derived key kind, however, implies that our current, basic primary key allocation is leading teams away from this best practice. And so if we include a distinction of primary vs. derived, there is an argument to extend the library to generate derived keys.

HMACs are a special case in that these secrets are a kind of derived secret in which we will expressly encode the non-sensitive correlating id of the secret used to initialize the algorithm. The open question is whether encoding this key kind is important for things like security response. There are an arbitrary number of unique generated CASK HMAC secrets that incorporate a common correlating id, while there is only a single representation of a non-HMAC primary or derived key.

michaelcfanning:
This significant revision of CASK contains the following changes:

- Sensitive data key moved from its current location. As a result of this move, we restore reserved padding bits for sensitive component sizes of 16, 32 and 64 bytes. A 48-byte sensitive component aligns cleanly along a 3-byte boundary and has no padding.
- The timestamp is moved to follow the CASK signature. Minutes data is retained.
- The expiry is dropped from the core format. This data would be helpful to responders to encode, as by reading the key it would be possible to understand whether the key, as originally allocated was valid in some exposure timeframe. A key CASK principle is for the core format to publish data that is useful for static consumption and not tempting to evaluate at runtime. And so we drop the expiry as too tempting for runtime evaluation.
- The CASK key kind is eliminated. Various classes of key, primary key, derived, HMAC can be specified in the provider-specific key kind component.
- We now encode the optional data size, as a count of 3-byte optional segments. By encoding this length as well as the sensitive component length in characters at a fixed location relative to the CASK signature `QJJQ` we enable high-performance detections that start with locating the CASK signature in test/data streams.
- The correlating id is dropped to 15 bytes in length. In general, a 16-byte value (128 bits) is preferred in scenarios for a strong guarantee of uniqueness. In practice, 120 bits is sufficient.

The BNF is renamed to `CaskSecrets.md` to reflect that there's no longer a specific CASK key kind in play. The churn in the BNF was sufficient that rather than renaming first so that we could diff, I simply created a clean copy of the new BNF. But in case anyone finds it helpful to see the explicit diffs from previous, I've left an edited version of the old BNF behind to examine. This will be deleted, obviously, once everything is reviewed/refined/accepted.

This change resolved #35, #36 and #37.


docs/PrimaryKey.md
```diff
@@ -1,67 +1,83 @@
-# CASK 256-bit Primary Keys
+# CASK Secrets
```
michaelcfanning, line 1: >Secrets [](http://example.com/codeflow?start=7&length=7)

I left these diffs here in case it's helpful to anyone to try to diff from the previous! This file will get deleted next.

.editorconfig
```diff
@@ -1,7 +1,7 @@
 root = true
+charset = utf-8
```
michaelcfanning, line 2: >charset [](http://example.com/codeflow?start=0&length=7)

We appear to have a file with a UTF8 BOM that is triggering the issue below. We can work around this for now by setting UTF8 as a default encoding (which means our mixture of UTF8 BOM + ASCII works). Alternately we could chase down the BOM and zap it (it may be in a project file, the issue below references BOMs being inserted in some project scenarios).

[https://github.com/dotnet/sdk/issues/46780](https://github.com/dotnet/sdk/issues/46780)

src/Cask/Cask.cs
```diff
@@ -28,13 +28,22 @@ public static bool IsCask(string key)
     /// <param name="key"></param>
     public static bool IsCask(ReadOnlySpan<char> key)
     {
-        if (key.Length < MinKeyLengthInChars || key.Length > MaxKeyLengthInChars || !Is4CharAligned(key.Length))
+        if (!IsValidKeyLengthInChars(key.Length))
+        {
+            return false;
+        }
+
+        SensitiveDataSize size = InferSensitiveDataSizeFromCharLength(key.Length);
```
michaelcfanning, line 36: >SensitiveDataSize [](http://example.com/codeflow?start=8&length=17)

This code is very close to some helper code in `CaskKey`, so should we share it? Helpers could be a place to do so, but the other API in helpers is very generic (i.e., not super CASK-y). Another approach would be an internal static on either this type or `CaskKey` and share the logic. Thoughts or other ideas?

src/Cask/CaskKey.cs
```diff
@@ -199,10 +183,4 @@ private static void ThrowFormat()
     {
         throw new FormatException("Input is not a valid Cask key.");
     }
-
-    [DoesNotReturn]
-    private static int ThrowUnrecognizedSensitiveDataSize(SensitiveDataSize sensitiveDataSize)
```
michaelcfanning, line 204: >ThrowUnrecognizedSensitiveDataSize [](http://example.com/codeflow?start=23&length=34)

I'd previously broken a core contract around this class, which is that if you manage to successfully create a `CaskKey` instance, you should be able to trust that the key does not include any invalid data. It is therefore not appropriate for exception handling of this kind (and some other removed logic).

src/Cask/Helpers.cs
```diff
@@ -43,94 +43,20 @@ public static bool Is4CharAligned(int charLength)
...
-                     "This is only meant to be called using the kind char of a known valid key.");
-        return (CaskKeyKind)(kindChar - 'A');
-    }
-
-    public static byte KindToByte(CaskKeyKind kind)
-    {
-        return (byte)((int)kind << CaskKindReservedBits);
-    }
 
-    public static byte ProviderKindToByte(char providerKind)
```
michaelcfanning, line 72: >ProviderKindToByte [](http://example.com/codeflow?start=23&length=18)

In general, this proposed update to the format looks like a nice simplification once rendered in code. Many helpers went away. The test pattern for decoding without any gee whiz construction logic looks acceptably clean/simple.

src/Cask/InternalConstants.cs
```diff
@@ -102,93 +100,19 @@ internal static partial class InternalConstants
     /// </summary>
     public const int ProviderKindReservedBits = 2;
 
-    /// <summary>
-    /// A bit mask to obtain the reserved bits from the key kind.
-    /// </summary>
-    public const int CaskKindReservedMask = (1 << CaskKindReservedBits) - 1;
```
michaelcfanning, line 108: >CaskKindReservedMask [](http://example.com/codeflow?start=21&length=20)

I've deleted a ton of constants and some exception helpers, based on code coverage and the general thought that anything that exists should have an explicit utilization in code.

I am not sure on the best way to replace our index/range constants. After the design update, all of this stuff needs to be computed dynamically. Basically, you first need to understand the size of the sensitive data component. This gives you the offset of the fixed cask signature, from which you can compute all remaining offsets.

src/Cask/Polyfill.cs
```diff
@@ -119,39 +119,11 @@ public static void ThrowIfNull([NotNull] object? argument, [CallerArgumentExpres
             }
         }
 
-        public static void ThrowIfGreaterThan(int value, int max, [CallerArgumentExpression(nameof(value))] string? paramName = null)
```
michaelcfanning, line 122: >ThrowIfGreaterThan [](http://example.com/codeflow?start=27&length=18)

Again, if we don't call it, I'm deleting it. I assume that we will naturally reconstruct these lovely pattern when there is a need to (perhaps as soon as when someone goes and adds timestamp validation, which I've left out of this PR simply due to time/size of change concerns)

src/Tests/Cask.Tests/ICask.cs
```diff
@@ -25,8 +25,8 @@ public interface ICask
 
     string GenerateKey(string providerSignature,
```
michaelcfanning, line 26: >GenerateKey [](http://example.com/codeflow?start=11&length=11)

It's nice to no longer repeated parameter types (string followed by string in particular) to lead to unexpected ordering issues...

Here's a new FxCop rule: on observing any two sequential argument types which are the same, require that all call sites use named arguments to distinguish them (and raise visibility on calling mistakes).

michaelcfanning:
Yes, I copied this note into my bundle of comments that will support the spec draft, coming soon.

---
In reply to: [2711260568](https://github.com/microsoft/cask/pull/43#issuecomment-2711260568) [](http://example.com/codeflow?ancestors=2711260568)

docs/CaskSecret.md
```diff
@@ -0,0 +1,83 @@
+# CASK Secrets
+## Standard Backus-Naur Form (BNF)
+```
+<key> ::= <sensitive-data> <cask-signature> <timestamp> <sensitive-data-size> <optional-data-size> <provider-kind> [<optional-fields>] <provider-signature> <correlating-id>
```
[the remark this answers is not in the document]
michaelcfanning: Yes, we could do this. The unfortunate rough edge of the format is that there is a single element of the timestamp that leaks across the next 4-char encoded boundary. 

Today, YMDHM precedes the next three encoded chars representing sensitive data size, optional data size and provider key kind. We made this choice because it effectively puts the provider key kind as a prefix to the provider signature, and that's very nice for regexes, other evaluation. So `BAZST`, `CAZST`, etc.

If we just said, well, let's do both, then we could have sensitive data size, optional data size, year and month in the first 3-byte segment, then day, hour, minute, provider key kind in the second. This PR already shows that these 6 bytes are best processed as a group, and the pattern here (I think) would be persistent even if we took this change, you'd just adjust the ordering of the various encoded/decoded char values. 

Here's a bunch of blah blah and a proposal on this change, see what you think.

[https://github.com/microsoft/cask/issues/46](https://github.com/microsoft/cask/issues/46)

docs/CaskSecret.md
```diff
@@ -0,0 +1,83 @@
...
+<provider-signature> ::= 4 * <base64url> ; Provider identifier (24 bits).
+<optional-fields> ::= { <optional-field> } ; Zero or more 4-character (24-bit) sequences of optional data.
+<optional-field> ::= 4 * <base64url> ; Each optional field is 4 characters (24 bits). This keeps data
+                                     ; cleanly aligned along 3-byte/4-encoded character boundaries,
+                                     ; facilitating readability of encoded form as well as byte-wise use.
+<correlating-id> ::= 20 * <base64url> ; 120 bits of unique, non-sensitive data generated by a cryptographically
+                                      ; secure RNG. This data is not sensitive and is designed to correlate 
+                                      ; metadata for generated secrets with reports of exposure and other data.
+                                      ; As a non-security-sensitive component, this data should never be used
+                                      ; to drive security-sensitive operations.
```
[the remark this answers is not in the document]
michaelcfanning: Yes, excellent, I had already changed this to `must never`, but `MUST NOT` is what we need, indeed. Thanks for teasing out this subtlety, it was really important because we hadn't previously properly appreciated all the implications of the design change.

docs/CaskSecret.md
```diff
@@ -0,0 +1,92 @@
...
+                   | <384-bits> | <512-bits-padded>                         ; such as random data generated by a cryptographically secure random
+                                                                            ; number generator (RNG), a Hash Message Authentication Code (HMAC),
+                                                                            ; an output of a Key Derivation Function (KDF), etc. CASK specifies
+                                                                            ; a storage location and component size for this data but does not
+                                                                            ; specify a particular cryptographic algorithm or method for
+                                                                            ; generating it. The size of this component must conform to the
+                                                                            ; encoded <sensitive-data-size> value.
+<128-bits-padded>  ::= 21 * <base64url> <base64-four-zeros-suffix> 'AA'     ; The total random data comprises 128 bits encoded as 21
```
nguerrera, line 22:
```suggestion
<128-bits-padded>  ::= 21 * <base64url> <base64-four-zeros-suffix> 'AA'     ; The total sensitive data comprises 128 bits encoded as 21
```

nguerrera: I'm not going to comment on all of them, but there are several more occurrences of `random` that should be `sensitive`.

src/Cask/Cask.cs
```diff
@@ -113,81 +174,79 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
...
         Debug.Assert(keyLengthInBytes <= MaxKeyLengthInBytes);
         Span<byte> key = stackalloc byte[keyLengthInBytes];
 
         // Entropy comprising the sensitive component of the key.
-        FillRandom(key[..SecretEntropyInBytes]);
-
-        // Sensitive component size.
-        key[SecretEntropyInBytes] = (byte)SensitiveDataSize.Bits256;
+        Index endOfSensitiveComponent = sensitiveDataSizeInBytes - intPaddingBytes;
+        FillRandom(key[..endOfSensitiveComponent]);
```
nguerrera, line 245: Apparently, the explicit writing of padding was removed in the previous change. I was surprised to not see it here.

We must not assume that stackalloc'ed memory is zero-ed out. 

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc

> The content of the newly allocated memory is undefined.

I think in practice with today's compilers, it will be zeroed out unless `SkipLocalsInit` is used, but we should not rely on this undocumented behavior.

src/Cask/Cask.cs
```diff
@@ -113,81 +174,79 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
...
-        return CaskKey.Encode(key);
-    }
-
-    private static void FinalizeKey(Span<byte> key,
-                                    int expiryInFiveMinuteIncrements,
-                                    ReadOnlySpan<char> providerData)
-    {
-        int bytesWritten;
+        Range caskSignatureByteRange = ComputeSignatureByteRange(sensitiveDataSize);
+        CaskSignatureBytes.CopyTo(key[caskSignatureByteRange]);
```
nguerrera, line 249: I am thinking that it would be easier to follow if generate did not compute ranges, but instead bumped a destination span like a pointer to write each field. 

I had done it this way originally but removed it when everything was constant, but now I regret that.

Something like this:

```
Span<byte> destination = key;

// Sensitive data.
FillRandom(destination[..SecretEntropyInBytes]);
destination = destination[SecretEntropyInBytes..];

// Padding.
destination[..paddingBytes].Clear();
destination = destination[paddingBytes..];

// Signature
CaskSignatureBytes.CopyTo(destination[..3]);
destination = destination[3..];

// etc.
```

Now I don't need to compute the signature range, just know that I will be positioned at the signature after writing the fields that come before it. This follows the pseudo-code (although that isn't update here) that just says "write X to the next N bytes".

nguerrera (commented):
I have not a completed a full review pass yet. Saving my comments so far. There are some high-level things that I think are worth discussing and I don't think that discussion needs to wait for me to review all the details.

src/Cask/Helpers.cs
nguerrera: I think tiny helpers that do things like `x - 'A' ` (and casts) are still applicable and would still be nice to use. I find comments like `// 'A' == index 0 of all printable base64-encoded characters.` in a larger method like `IsCaskBytes` to be a hint to extract a helper, same as seeing `15` as a hint to extract a constant.

nguerrera (commented):
Still reviewing, but saving progress.

src/Cask/SensitiveDataSize.cs
```diff
@@ -4,26 +4,36 @@
...
+/// values below map both to the base64url-encoded index of the value's printable
+/// character, as well as a numeric value that is the number of 16-byte segments
+/// of the sensitive data component.
 /// </summary>
 public enum SensitiveDataSize
 {
+    None = 0,
+
+    /// <summary>
+    /// Specifies a computed value with one 16-byte
```
nguerrera, line 17: Super nit that you can ignore, but this is quite a narrow word-wrapping at column 52.

FWIW, I use https://marketplace.visualstudio.com/items?itemName=stkb.rewrap and ALT+Q to wrap to its default width of 80.

src/Cask/SensitiveDataSize.cs
```diff
@@ -4,26 +4,36 @@
...
+    /// (128 bit) segment of entropy.
```
nguerrera, line 18: s/entropy/sensitive data/

Many instances of this in this file, but I will only comment once.

src/Tests/Cask.Tests/CaskKeyTests.cs
```diff
@@ -7,26 +7,30 @@
...
 {
-    [Fact]
-    public void CaskKey_UninitializedKindAccessThrows()
-    {
-        CaskKey key = default;
-        Assert.Throws<InvalidOperationException>(() => key.Kind);
-    }
+    internal static SensitiveDataSize[] AllSensitiveDataSizes => [
+        SensitiveDataSize.Bits128, SensitiveDataSize.Bits256,
+        SensitiveDataSize.Bits384, SensitiveDataSize.Bits512];
```
nguerrera, line 19: nit, I'd put each of these on their own line.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -22,39 +24,43 @@ protected CaskTestsBase(ICask cask)
...
-    [InlineData("Y7G_WqVrIxJ9y3kqLdX6OOhTwC1kTF0eWQidLckLqfEAQJJQTESTMPAlrkxagZHvE1rmbBnVwEHZBBRVnAAA_NG_", CaskKeyKind.PrimaryKey)]
-    [InlineData("V5ja_SGw4_eyqKw-mBfx8DlqjJfea4Qs5B6AR3HjlgwAQJJQTESTMPCK8K_4JYG3ppYTmdnSS4TcBBQXDAAA_NG_", CaskKeyKind.PrimaryKey, "CK8K_4JYG3ppYTmdnSS4Tc")]
-    public void CaskSecrets_EncodedMatchesDecoded(string encodedKey, CaskKeyKind expectedKeyKind, string expectedC2Id = "")
-    {
-        TestEncodedMatchedDecoded(encodedKey, expectedKeyKind, expectedC2Id);
-    }
-
-    [Fact]
-    public void CaskSecrets_EncodedMatchesDecoded_GeneratedKey()
+    [InlineData("", SensitiveDataSize.Bits128), InlineData("-MF--NG--RW--RG-", SensitiveDataSize.Bits128)]
```
nguerrera, line 44: Who is RG? 😊

.editorconfig
nguerrera: I will look into this.

nguerrera (commented):
OK, I've made a full review pass now.

.editorconfig
michaelcfanning: That baton is yours.

docs/CaskSecret.md
michaelcfanning: yes thanks

src/Cask/CaskKey.cs
michaelcfanning: No worries, code is right now and my mental model is updated.

src/Cask/Helpers.cs
michaelcfanning: Agree! Pending future PR.
[https://github.com/microsoft/cask/issues/47](https://github.com/microsoft/cask/issues/47)

src/Cask/InternalConstants.cs
michaelcfanning: Noted in #47

src/Cask/Cask.cs
```diff
@@ -63,6 +68,60 @@ public static bool IsCask(ReadOnlySpan<char> key)
...
+
+        int lengthInBytes = lengthInChars / 4 * 3;
+        return InferSensitiveDataSizeFromByteLength(lengthInBytes);
+    }
+
+    internal static SensitiveDataSize InferSensitiveDataSizeFromByteLength(int lengthInBytes)
+    {
+        /* 
+         *  Required CASK encoded data, 30 bytes.
+         *  QJJQ YMDH MLOP TEST 1234 1234 1234 1234 1234 1234
```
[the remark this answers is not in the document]
michaelcfanning: Woof, yes, this area moved around a little. Fixed, thanks.

src/Cask/Cask.cs
```diff
@@ -63,6 +68,60 @@ public static bool IsCask(ReadOnlySpan<char> key)
...
+         * 
+         *  128-bit : 45 bytes (18 bytes sensitive + 27 reserved) : 12 bytes of optional data permissible < (60 - 45)
+         *  256-bit : 60 bytes (33 bytes sensitive + 27 reserved) : 12 bytes of optional data permissible < (75 - 60)
+         *  384-bit : 75 bytes (48 bytes sensitive + 27 reserved) : 15 bytes of optional data permissible < (93 - 75)
+         *  512-bit : 93 bytes (66 bytes sensitive + 27 reserved) : 15 bytes (value chosen to align with 384 bit keys)
```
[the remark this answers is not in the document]
michaelcfanning: OK, so my reply is that we enforce a 12 byte constraint today but that, if true, that means there is a specific 384-bit key in code that's forbidden where we have space for it. Let me go add a test and look for that bug,.

I've gone back and forth: should we permit the extra 3 bytes in 384 just because we can? if we do that, we should we allow 15 bytes in the longer key because it's weird to drop back the permissible optional data length for this one? or should we just say, 12 bytes only because that's simplest to explain?

I will add explicit tests that ensure we cover our intent thoroughly. Please let me know what you think that intent should be.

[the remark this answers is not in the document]
nguerrera: 12 bytes max, period. That's my preference. 😊

src/Cask/Cask.cs
```diff
@@ -63,6 +68,60 @@ public static bool IsCask(ReadOnlySpan<char> key)
...
+
+    internal static SensitiveDataSize InferSensitiveDataSizeFromCharLength(int lengthInChars)
+    {
+        Debug.Assert(IsValidKeyLengthInChars(lengthInChars));
+
+        int lengthInBytes = lengthInChars / 4 * 3;
+        return InferSensitiveDataSizeFromByteLength(lengthInBytes);
+    }
+
+    internal static SensitiveDataSize InferSensitiveDataSizeFromByteLength(int lengthInBytes)
```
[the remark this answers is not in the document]
michaelcfanning: More thoughts than you wanted on this idea, let's discuss after this review.
[https://github.com/microsoft/cask/issues/48](https://github.com/microsoft/cask/issues/48)

src/Cask/Cask.cs
```diff
@@ -73,8 +132,11 @@ public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUtf8)
             return false;
         }
 
-        // Check for CASK signature, "QJJQ" UTF-8 encoded.
-        if (!keyUtf8[CaskSignatureCharRange].SequenceEqual(CaskSignatureUtf8))
+        SensitiveDataSize sensitiveDataSize = InferSensitiveDataSizeFromCharLength(keyUtf8.Length);
+        Range caskSignatureByteRange = ComputeSignatureCharRange(sensitiveDataSize);
```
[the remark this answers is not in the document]
michaelcfanning: right! `caskSignatureUtf8Range` is longer but would be consistent with other names.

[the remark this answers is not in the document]
michaelcfanning: `caskSignatureUtf8Range`

src/Cask/Cask.cs
```diff
@@ -113,81 +174,79 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
...
-            kind is not CaskKeyKind.PrimaryKey and not CaskKeyKind.DerivedKey and not CaskKeyKind.HMAC)
+        int sensitiveDataSizeInBytes = RoundUpTo3ByteAlignment((int)encodedSensitiveDataSize * 16);
+        int expectedKeyLengthInBytes = sensitiveDataSizeInBytes + FixedKeyComponentSizeInBytes + providerDataLengthInBytes;
+        if (expectedKeyLengthInBytes != keyBytes.Length)
         {
             return false;
         }
 
-        // TBD: we should author a TryBytesToTimestampAndExpiry method.
+        // TODO: Review Cask.IsCaskBytes and its callers carefullyto ensure all useful checks are made.
```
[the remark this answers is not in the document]
michaelcfanning: thanks for the fix

src/Cask/Cask.cs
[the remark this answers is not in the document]
michaelcfanning: ok! I will make it so. I can't resist noting here, that if we want the simplification of limiting all optional data to 12 bytes, that opens the possibility of us injecting 3 bytes of padding into the 384 bit key. the result of that is that we would *always* have at least 8 bits of reserved space in the format for some future use. it comes at the cost of artificially creating padding for the 384 bit key unnecessarily. that key is already long, the space wouldn't kill us. but adding this padding would be an unnecessary additional wrinkle/requirement we could avoid. thoughts?

src/Cask/Cask.cs
michaelcfanning: Oh geez, ok, what could go wrong with undefined initialization behavior, it's worked out great for C++. :)

Haha. The previous explicit padding was removed when I adjusted this code for optional data sizes. I added this in the test decoding but not here, will do it.

docs/CaskSecret.md
```diff
@@ -0,0 +1,92 @@
+# CASK Secrets
```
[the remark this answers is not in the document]
michaelcfanning: Thank you, yes, done.

src/Cask/Cask.cs
michaelcfanning: Useful, let's do this when we fix #47.
[https://github.com/microsoft/cask/issues/49](https://github.com/microsoft/cask/issues/49)

src/Cask/Cask.cs
```diff
@@ -63,6 +68,60 @@ public static bool IsCask(ReadOnlySpan<char> key)
         return IsCaskBytes(keyBytes);
     }
```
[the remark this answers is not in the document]
michaelcfanning: I agree with this. I moved these helpers. I also verified the need for internal on anything marked as such and found several that could be dropped to private. I put the single internal helper at the top, followed by privates, followed by internal helpers strictly provided for mocking.

src/Cask/Helpers.cs
```diff
@@ -43,94 +43,20 @@ public static bool Is4CharAligned(int charLength)
...
-    {
-        int base64Index;
+        int keyLengthInBytes = providerDataLengthInBytes + FixedKeyComponentSizeInBytes;
 
-        const int uppercaseZIndex = 'Z' - 'A';
-        const int lowercaseZIndex = 'z' - 'a';
+        int entropyInBytes = (int)sensitiveDataSize * 16;
+        int sensitiveDataSizeInBytes = RoundUpTo3ByteAlignment(entropyInBytes);
```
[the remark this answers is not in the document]
michaelcfanning: Fixed everywhere. Also changed many names from 'sensitiveDataSize' to 'secretSize'. Arg. Why did I include that rename here? :P

src/Cask/Helpers.cs
```diff
@@ -43,94 +43,20 @@ public static bool Is4CharAligned(int charLength)
...
-    /// Returns false if the reserved bits in that byte are non-zero.
-    /// </summary>
-    public static bool TryByteToKind(byte value, out CaskKeyKind kind)
-    {
-        if ((value & CaskKindReservedMask) != 0)
-        {
-            kind = default;
-            return false;
-        }
+        keyLengthInBytes += sensitiveDataSizeInBytes;
```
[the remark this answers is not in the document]
michaelcfanning: simplified.

src/Cask/Limits.cs
```diff
@@ -13,40 +13,36 @@ namespace CommonAnnotatedSecurityKeys;
  * it's possible that the JIT will not discover that the computation yields a constant on
  * every invocation, but it will treat static readonly fields as constants.
  */
+
```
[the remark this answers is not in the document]
michaelcfanning: Yes, that's why I did it. SEV4. tell me what to do if anything.

src/Cask/SensitiveDataSize.cs
michaelcfanning: This is just me being precious, I 'columnized' this data to match the literal code and online comments following.

It's much better for the team to have a consistent, tool-driven approach. I've installed rewrap and will use it moving forward. It's a lot faster than hand-authoring these, wowza.

src/Tests/Cask.Tests/CaskSecretsTests.cs
michaelcfanning: A key friend of ours on a certain review board who has contributed essential thoughts to the effort.

docs/GenerateKeyPseudoCode.md
```diff
@@ -14,44 +14,42 @@
...
     - Provider signature is exactly 4 characters long.
-    - Provider signature consists entirely of characters that are valid in base64url encoding.
-    - Provider key kind is a single, printable (i.e., non-padding) base64url character.
-    - Expiry in five minutes increments is a non-negative integer less than 262,144.
-    - Provider data (if non-empty) has a length that is a multiple of 4 characters and no more than 32 characters.
+    - Provider signature consists entirely of printable (non-padding) characters that are valid in base64url encoding.
+    - Provider key kind is a single, printable base64url character.
+    - Provider data (if non-empty) has a length that is a multiple of 4 characters and no more than 12 characters.
     - Provider data (if non-empty) consists entirely of base64url printable characters.
+    - Secret data size is between 1 (indicating a 128-bit component) and 4 (indicating 512-bits).
```
nguerrera, line 21: Not listed as an input.

nguerrera, line 21: We can discuss separately from this PR, but I think we should consider making this a size in bytes or bits. And then specify the required alignment and min/max to validate. If someone decided to implement this without an enum, it would be really weird and hard to use IMHO.

docs/GenerateKeyPseudoCode.md
```diff
@@ -14,44 +14,42 @@
...
 1. Let N = the length of the base64url-decoded provider data.
     - Number of characters in provider data divided by 4, times 3.
+1. Compute the sensitive data size in bytes:
+    - Multiply the secret data size by 16.
+    - Round this value up to the next multiple of 3, if necessary.
```
nguerrera, line 26: You can leave this for now, but I'm finding it hard to understand that this does not mean to discard the unrounded value, that you actually do need later. Maybe it would be better to specify how to compute how many bytes to pad separately.

docs/GenerateKeyPseudoCode.md
```diff
@@ -14,44 +14,42 @@
...
+1. Retrieve the current date and time in UTC and store the result in T.
 1. Encode T in 4 characters, YMDH:
-    - Y = base64url-encoding of (Year - 2025).
-    - M = base64url-encoding of zero-based month.
-    - D = base64url-encoding of zero-based hour.
-    - H = base64url-encoding of zero-based day.
+    - Y = base64url-encoding of T.Year - 2025.
+    - M = base64url-encoding of T zero-based month.
+    - D = base64url-encoding of T zero-based hour.
+    - H = base64url-encoding of T zero-based day.
```
nguerrera, line 43: Looks like there's an unfinished idea here on `T.` syntax. Revert it or finish it :) Or maybe `T.Year` is a typo? But `T zero-based month` doesn't really parse for me.

docs/GenerateKeyPseudoCode.md
```diff
@@ -14,44 +14,42 @@
...
 1. Base64url-decode YMDH and store the result in the next 3 bytes.
-1. Base64url-encode T minutes M in a single character
-1. Encode the last 3 bytes of the big-endian representation of the 18-bit expiry.
-1. Base64url-decode M and the 3-character expiry and store the result in the next 3 bytes.
-1. Base64url-decode provider data and store the result in the next N bytes.
+1. Encode the timestamp minutes, data sizes and provider key kind in 4 characters, MSOK:
+    - M = base64url-encoding of T zero-based minute.
```
nguerrera, line 46: Note to make this consistent with any changes for comment above about `T.`.

nguerrera (commented):
Review of pseudo-code only. Now reviewing the rest since last review.

src/Cask/Cask.cs
```diff
@@ -287,6 +205,108 @@ public static CaskKey GenerateKey(string providerSignature,
...
+
+    private static SecretSize InferSensitiveDataSizeFromCharLength(int lengthInChars)
+    {
+        Debug.Assert(IsValidKeyLengthInChars(lengthInChars));
+
+        int lengthInBytes = lengthInChars / 4 * 3;
+        return InferSensitiveDataSizeFromByteLength(lengthInBytes);
+    }
+
+    private static SecretSize InferSensitiveDataSizeFromByteLength(int lengthInBytes)
```
nguerrera, line 232: nit: rename here too?
```suggestion
    private static SecretSize InferSecretSizeFromByteLength(int lengthInBytes)
```

nguerrera (approved):

docs/GenerateKeyPseudoCode.md
michaelcfanning: Nuts and expiry not deleted. Let's fix this before checking in.

michaelcfanning: More accurately, this value is a count of 16-byte segments in the secret data. Just as our optional data value is a count of 3-byte segments. I updated the language. Glad to continue discussing this.

nguerrera (approved):

PR #54, opened 2025-03-24
Convert to destination span pattern.

michaelcfanning:


src/Cask/Cask.cs
```diff
@@ -100,7 +100,7 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
         }
 
         int providerDataLengthInBytes = (minutesSizesAndKeyKindChars[2] - 'A') * 3;
-        if (providerDataLengthInBytes > MaxProviderDataLengthInBytes || !Is3ByteAligned(providerDataLengthInBytes))
```
michaelcfanning, line 103: >Is3ByteAligned [](http://example.com/codeflow?start=73&length=14)

Whoops, this code could never fail given that we just assigned a value to it guaranteed to be a multiple of 3.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -570,11 +736,19 @@ private void IsCaskVerifySuccess(string key)
         Assert.True(Cask.IsCaskBytes(keyBytes), $"'IsCask(byte[])' failed for: {key}'.");
     }
 
-    private void IsCaskVerifyFailure(string key)
+    private void IsCaskVerifyFailure(string key, bool expectedIsMatchResult = false)
```
nguerrera, line 739: It took me a minute to understand that this parameter relates to the regex match. My brain seemed to think IsMatch was IsCask. Consider `bool expectedRegexIsMatchResult` or something shorter that includes regex.

SEV 4.

nguerrera (approved):

src/Tests/Cask.Tests/CaskSecretsTests.cs
michaelcfanning: Sure! I like this SEV4.

Issue #49, opened 2025-03-12
Clean up secret construction by bumping a destination span


michaelcfanning:
Our CASK secret construction and decomposition is easily accomplished establishing the padded sensitive component size and moving forward through the key.  We should clean up the code to bump a destination span to perform our work. This approach in concert with descriptive local names will make the code readable.

This issue probably should be completed in concert with #47.

``` 
Span<byte> destination = key;

// Sensitive data.
FillRandom(destination[..SecretEntropyInBytes]);
destination = destination[SecretEntropyInBytes..];

// Padding.
destination[..paddingBytes].Clear();
destination = destination[paddingBytes..];

// Cask signature.
CaskSignatureBytes.CopyTo(destination[..3]);
destination = destination[3..];

// etc.
```

michaelcfanning:
This issue resolved by #54.

Issue #48, opened 2025-03-11
Should we prefer a descriptive key header to avoid the non-obvious complexity of determining key lengths?


michaelcfanning:
Determining key lengths while maintaining efficient detection of keys in both textual and data streaming scenarios. 

CASK allows for two kinds of optionally sized data, a sensitive data component and a provider-specified sequence of 3-byte segments allowing for 0 - 12 bytes of data.

That artificial limit on optional size allows us to play a trick in that the legal minimal and maximal ranges of, say, a 128-bit are known in advance and these sizes don't overlap with any other legal range for other sensitive data sizes. Stated more simply, a 128-bit CASK secret's length can never exceed the length of any larger key.

These constraints make validity checks more efficient and allow us to create more efficient/constraint regular expressions for detecting these patterns.

Despite this capability, we *still* encode sensitive and optional data sizes in the key itself at the cost of consuming 12 bits of data (two encoded characters). The reason is to enable high-performance detection in non-delimited, data streaming services. For these, a matcher can process data at high speeds until locating the fixed CASK signature `QJJQ`. On finding this data, the encoded lengths allow seeking forward and backwards to validate the detection.

All of these hijinks could be avoided by simply encoding some or all descriptive metadata at the very beginning or the very end of the key. In general, we have avoided expressing fixed patterns at the beginning of secrets to discourage the possibility of humans observing this consistency and subverting security controls simply by eliding a fixed prefix.

For CASK, however, we have dynamically generated content (specifically the timestamp component) that could be sufficient to serve the purpose of providing sufficient friction to deter a lazy actor.

Today, the CASK signature `QJJQ`, with all non-sensitive data following it, provides a nice property: a user can easily grab the signature and all data to the right of it and be confident no sensitive information is included. This value would be lost with a secret header, as there would be no obvious transition from the provider-specific data into the sensitive component. i.e., the CASK signature is the only stable anchoring pattern in our format. We could arguably artificially inject a second 'terminating' signature, such as a sequence of hyphen `----` to indicate transition into the sensitive data, at the expense of four additional encoded characters across the wire.

michaelcfanning:
[...]

Issue #45, opened 2025-03-10
Review `Cask.IsCaskBytes` and its callers carefully to ensure all useful checks are made


michaelcfanning:
And that, where possible, duplicative checks do not occur.

PR #55, opened 2025-03-30
Reorder sensitive data, optional data size, et al. as described in #48

### LINKED ISSUE #46: Consider re-ordering sensitive data size, et al. in the `timestampSizesAndKind` 6-byte data sequence.
The current proposal orders this data as so, in a sequence of 8 encoded characters, that are prefixed the with `QJJQ` fixed signature and followed by the provider fixed signature:

Year/Month/Day/Hour/Minute/Sensitive Data Size/Optional Data Size/Provider Key Kind

It turns out in practice that handling this data range is best accomplished by encoding its bytewise representation (if that's what you have) and processing each encoded 6-bit character on its own. The specific ordering of this data doesn't matter, if we accept this approach.

A better (possibly simply overthought) representation would encode the sensitive data size immediately following the CASK signature, so `QJJQB` would represent the universe of CASK 256-bit secrets, etc. Similarly, it's helpful to keep the provider key immediately preceding the provider signature, so that key sub-types like `ATEST`, `BTEST` etc. might provide easy visual inspection in exposed secrets, focused regexes for detecting them, etc. 

The optional data size could follow the sensitive data size, since conceptually they are similar and on parsing this data, all facts re: secret length are known. *or* another argument could be made to co-locate this data value close to the key provider. 

Personally, I think I vote for the following (though I don't have strong feelings). Sensitive data is right next to the CASK, followed by year/month/day of allocation. Optional data size, provider key kind precede the provider signature.

Sensitive Data Size/Year/Month/Day/Hour/Minute/Optional Data Size/Provider Key Kind

michaelcfanning:
This PR establishes a new ordering of the raw 6-byte sequence that incorporates the timestamp, sensitive data size, optional data size, and provider key kind. Previously the timestamp preceded all other data, so the practical description of the required change is to move this data to precede the provider kind (leaving the sensitive data and optional data sizes in the same order and now immediately following the CASK signature).

#46 includes the detailed rationale for the change). Each encoded character (of which there are 8) in this base64-encoded sequence is interpreted independently, as follows and now in the following order:

1. Secret data size
2. Optional data size
3. Timestamp (viz., time-of-key-allocation) year
4. Timestamp  month
5. Timestamp  day
6. Timestamp  hour
7. Timestamp  minute
8. Provider key kind.

In addition to updating the BNF, pseudocode and API logic, this change adds several missing tests to validate that illegal timestamp values are not accepted as valid.

This change resolves #46 and #45. I have also opportunistically continued to implement our destination `Span` pattern for accessing data (in places not already covered by #49).

### DIFF
--- a/docs/CaskSecret.md
+++ b/docs/CaskSecret.md
@@ -3,9 +3,9 @@
 ```
 <key> ::= <sensitive-data>      ; A sequence of security-sensitive bytes.
           <cask-signature>      ; A fixed signature (`QJJQ`) that enables high-performance textual identification.
-          <timestamp>           ; The year, month, day, hour, and minute of secret allocation.
           <sensitive-data-size> ; A count of 16-byte segments encoded as sensitive data ('B' = 1 x 16 bytes = 128 bits, etc).
           <optional-data-size>  ; A count of 3-byte optional data segments, 'A' = 0 = 0 bytes, 'B' = 1 = 3 bytes, etc.
+          <timestamp>           ; The year, month, day, hour, and minute of secret allocation.
           <provider-kind>       ; A provider-defined key kind.
           [<optional-fields>]   ; Optional fields comprising provider-defined data.
           <provider-signature>  ; A fixed signature identifying the secret provider.
@@ -41,14 +41,14 @@
                             | 'g' | 'k' | 'o' | 's' | 'w' | '0' | '4' | '8' ; Base64 printable characters with two trailing zero bits.
 <base64-four-zeros-suffix> ::= 'A' | 'Q' | 'g' | 'w'                        ; Base64 printable characters with four trailing zero bits.
 <cask-signature> ::= 'QJJQ'                                                 ; Fixed signature identifying the CASK key.
+<sensitive-data-size> ::= 'A'                                               ; 'A' indicates a 128-bit sensitive data component, 'B' 256-bit, etc.
+<optional-data-size> ::= 'A'                                                ; 'A' = zero 3-byte optional bytes, 'B' = one optional 3-byte segment, etc.
 <timestamp> ::= <year> <month> <day> <hour> <minute>                        ; Time-of-allocation timestamp components.
 <year> ::= <base64url>                                                      ; Allocation year, 'A' (2025) to '_' (2088).
 <month> ::= 'A'..'L'                                                        ; Allocation month, 'A' (January) to 'L' (December).
 <day> ::= 'A'..'Z' | 'a'..'e'                                               ; 'A' = day 1, 'B' = day 2, ... 'e' = day 31
 <hour> ::= 'A'..'X'                                                         ; Represents hours 0-23. 'A' = hour 0 (midnight), ... 'X' = hour 23.
 <minute> ::= 'A'..'7'                                                       ; Represents minutes 0-59.
-<sensitive-data-size> ::= 'A'                                               ; 'A' indicates a 128-bit sensitive data component, 'B' 256-bit, etc.
-<optional-data-size> ::= 'A'                                                ; 'A' = zero 3-byte optional bytes, 'B' = one optional 3-byte segment, etc.
 <provider-kind> ::= <base64url>                                             ; Provider-defined key kind.
 <provider-signature> ::= 4 * <base64url>                                    ; Provider identifier (24 bits).
 <optional-fields> ::= { <optional-field> }                                  ; Zero or more 4-character (24-bit) sequences of optional data.
@@ -68,8 +68,8 @@
 |decodedKey[..31]|0...255|0x0...0xFF|00000000b...11111111b|256 bits of sensitive data produced by a cryptographically secure RNG, an HMAC, etc.|
 |decodedKey[32]|0|0x00|00000000b| 8 bits of reserved padding.
 |decodedKey[33..36]| 37, 4, 9  |0x40, 0x92, 0x50| 00100000b, 10010010b, 01010000b | Decoded 'QJJQ' signature.
-|decodedKey[36..39]||||Timestamp data encoded in 4 six-bit segments for YMDH.
-|decodedKey[39..42]||||Timestamp minutes, sensitive data size, optional-data-size, and provider kind data encoded in 4 six-bit segments.
+|decodedKey[36..39]||||Sensitive data size, optional-data-size, timestamp year and month encoded in 4 six-bit segments.
+|decodedKey[39..42]||||Timestamp day, hour, minutes and provider kind data encoded in 4 six-bit segments.
 |decodedKey[42..45]|0...255|0x0...0xFF|00000000b...11111111b| Provider signature, e.g. , '0x4c', '0x44', '0x93' (base64-encoded as 'TEST')
 |decodedKey[45..60]||||16 byte non-sensitive, unique correlating id.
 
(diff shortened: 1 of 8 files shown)


src/Tests/Cask.Tests/CSharpCaskTests.cs
```diff
@@ -41,9 +41,9 @@ public bool IsCask(string key)
             }
             catch (FormatException)
             {
-                // On receiving this exception, we have invalid base64 input.
-                // As a result, we will skip the IsCaskBytes check, which
-                // will throw for this condition.
+                // On receiving this exception, we have invalid base64 input. As
```
michaelcfanning, line 44: >receiving [](http://example.com/codeflow?start=22&length=9)

Made some effort to reformat code comments using rewrap to get ourselves into its standard formatting condition.

src/Cask/Cask.cs
```diff
@@ -186,26 +217,26 @@ public static CaskKey GenerateKey(string providerSignature,
...
+            Base64UrlChars[now.Minute],      // Zero-index minute.
+            providerKeyKind,
         ];
 
-        bytesWritten = Base64Url.DecodeFromChars(chars, destination[..MinuteSizesAndKeyKindSizeInBytes]);
-        Debug.Assert(bytesWritten == MinuteSizesAndKeyKindSizeInBytes);
-        destination = destination[MinuteSizesAndKeyKindSizeInBytes..];
+        bytesWritten = Base64Url.DecodeFromChars(chars, destination[..DayHourMinutesKeyKindSizeInBytes]);
+        Debug.Assert(bytesWritten == DayHourMinutesKeyKindSizeInBytes);
+        destination = destination[DayHourMinutesKeyKindSizeInBytes..];
```
nguerrera, line 239: Pre-existing, but can we decode all 8 chars at once?

src/Cask/Cask.cs
```diff
@@ -126,23 +129,51 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
...
+            return false;
+        }
 
-        return true;
+        char minute = dayHourMinutesAndProviderKindChars[2];
+        // An encoded minute (a zero-indexed value) that exceeds 30 ('7') is not valid.
+        return
+            (minute >= 'A' && minute <= 'Z') ||
+            (minute >= 'a' && minute <= 'z') ||
+            (minute >= '0' && minute <= '7');
```
nguerrera, line 166: I still find the back-and-forth between base64 and bytes very confusing and very complicated.  But I think I've lost on sacrificing the readability of the timestamp to address that so I should really just stop commenting on it. 😊

src/Cask/Cask.cs
```diff
@@ -99,20 +99,23 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
...
-        Range ymdhTimestampRange = caskSignatureByteRange.End..(caskSignatureByteRange.End.Value + CaskSignatureSizeInBytes);
-        Range minutesSizesAndKeyKindRange = ymdhTimestampRange.End..(ymdhTimestampRange.End.Value + YearMonthDayHourSizeInBytes);
+        ReadOnlySpan<byte> destination = keyBytes[caskSignatureByteRange.End.Value..];
 
-        Span<char> minutesSizesAndKeyKindChars = stackalloc char[4];
-        int bytesWritten = Base64Url.EncodeToChars(keyBytes[minutesSizesAndKeyKindRange], minutesSizesAndKeyKindChars);
+        ReadOnlySpan<byte> secretOptionalSizesYearMonthBytes = destination[..DataLengthsYearMonthSizeInBytes];
+        destination = destination[DataLengthsYearMonthSizeInBytes..];
+
+        Span<char> secretOptionalSizesYearMonthChars = stackalloc char[DataLengthsYearMonthSizeInChars];
```
nguerrera, line 107: Flip-side of my last question here: can we encode all 8 chars at once?

src/Cask/CaskKey.cs
```diff
@@ -172,7 +172,10 @@ private void ThrowIfNotInitialized()
     }
 
     // language=regex
-    private const string RegexPattern = """(^|[^A-Za-z0-9+\/\-_])([A-Za-z0-9\-_]{4}([A-Za-z0-9\-_]{20}){1,3}|[A-Za-z0-9\-_]{88})?QJJQ[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][B-E][A-E][A-Za-z0-9\-_]([A-Za-z0-9\-_]{4}){0,4}[A-Za-z0-9\-_]{24}([^A-Za-z0-9+\/\-_]|$)""";
+    private const string RegexPattern =
+        """
+        (^|[^A-Za-z0-9+\/\-_])([A-Za-z0-9\-_]{24}QJJQB|[A-Za-z0-9\-_]{44}QJJQC|[A-Za-z0-9\-_]{64}QJJQD|[A-Za-z0-9\-_]{88}QJJQE)(A[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{24}|B[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{28}|C[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{32}|D[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{36}|E[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{40})([^A-Za-z0-9+\/\-_]|$)
+        """;
```
nguerrera, line 178: Pre-existing regex issue, I think: we're not providing a way to extract just the match and not include the delimiting characters.

docs/CaskSecret.md
```diff
@@ -41,14 +41,14 @@
                             | 'g' | 'k' | 'o' | 's' | 'w' | '0' | '4' | '8' ; Base64 printable characters with two trailing zero bits.
 <base64-four-zeros-suffix> ::= 'A' | 'Q' | 'g' | 'w'                        ; Base64 printable characters with four trailing zero bits.
 <cask-signature> ::= 'QJJQ'                                                 ; Fixed signature identifying the CASK key.
+<sensitive-data-size> ::= 'A'                                               ; 'A' indicates a 128-bit sensitive data component, 'B' 256-bit, etc.
+<optional-data-size> ::= 'A'                                                ; 'A' = zero 3-byte optional bytes, 'B' = one optional 3-byte segment, etc.
```
nguerrera, line 45: Pre-existing, but why are these `::= 'A'`? Shouldn't they be `::= <base64url>`?

docs/CaskSecret.md
```diff
@@ -3,9 +3,9 @@
 ```
 <key> ::= <sensitive-data>      ; A sequence of security-sensitive bytes.
```
nguerrera, line 4: Pseudo-code also need updating.

src/Cask/CaskKey.cs
```diff
@@ -172,7 +172,10 @@ private void ThrowIfNotInitialized()
     }
 
     // language=regex
-    private const string RegexPattern = """(^|[^A-Za-z0-9+\/\-_])([A-Za-z0-9\-_]{4}([A-Za-z0-9\-_]{20}){1,3}|[A-Za-z0-9\-_]{88})?QJJQ[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][B-E][A-E][A-Za-z0-9\-_]([A-Za-z0-9\-_]{4}){0,4}[A-Za-z0-9\-_]{24}([^A-Za-z0-9+\/\-_]|$)""";
+    private const string RegexPattern =
```
nguerrera, line 175: Just leaving a link to the follow-up item you filed: https://github.com/microsoft/cask/issues/57 from the relevant line of this PR.

docs/CaskSecret.md
nguerrera: Description says it was updated, but I don't see it in the diff. Maybe it wasn't pushed?

src/Cask/Cask.cs
michaelcfanning: Yes, will fix as clean-up in the next change if that's ok. I'm going to go update to check month and day combinations for validity. So will be working roughly in this area.

src/Cask/Cask.cs
michaelcfanning: Haha. Even if you argued for an opaque, bytewise timestamp, we'd have complexity for handling the data preceding and following it. :)

src/Cask/CaskKey.cs
michaelcfanning: Gee, ok. Any suggestion on a name? we use ?<secret> in some contexts and ?<refine>. Honestly, I'd guess we could simply eliminate the prefix/suffix requirement. I don't think I captured the precise expected collision rate with our format and randomly generated data and may need to reproduce it.

If we eliminate the suffix and prefix entirely, then you need to increase the odds of collision by change for each character in the 'sliding window' of potential matches.

e.g., if you try to match a single character (as delimited in our regex), the odds are 1 in 64 of a collision. if your data sample is 64 characters, with no delimiter, the chances are even.

so the question is, given the very small likelihood of collision, what length of unbroken base64 encoded data would result in a meaningful chance of collision that argues for delimiters + capture group?

docs/CaskSecret.md
michaelcfanning: oops will fix, thanks for catching this debt.

src/Cask/CaskKey.cs
nguerrera: Let's chat about this one separately. I don't think I understand this and since it isn't something changed by this PR, it will be easier to talk about it in isolation.

nguerrera (approved):

src/Cask/CaskKey.cs
michaelcfanning: So, for the 384-bit scenario, odds of collision are 1 in 107,612,894,321 for any sequence of appropriate size. If we create a sliding window (i.e., regex with no delimiters) processing roughly 50 GB of streaming base64 (URL-safe, s/be noted) data would have a 50% chance of producing a CASK false positive.

So on surface, the delimiters in the regex allow for reasonable collision rates with this solution only. If we want a streaming scenario, we could reintroduce a partial checksum to eliminate FPs (for the general CASK detection. Introducing the provider signature for any specific check will add another 2 ^ 24 (1 in 16.77M) multiplier to odds..

PR #61, opened 2025-04-08
Target netstandard2.0 instead of net472, remove System.IO.Hashing dependency

nguerrera:
The library compiles fine for netstandard2.0 now, so use that in place of net472. Benchmarks and tests still use net472 to exercise the netstandard2.0 compilation. This aligns with https://github.com/security-utilities, which will make it easier to move code between these repos.

Also remove the System.IO.Hashing that we no longer use.

### DIFF
--- a/src/Cask/Cask.csproj
+++ b/src/Cask/Cask.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net472</TargetFrameworks>
+    <TargetFrameworks>net8.0;netstandard2.0</TargetFrameworks>
     <IsPackable>true</IsPackable>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


nguerrera:
This is our dependency graph now:


```
net8.0
    Microsoft.Bcl.Memory

netstandard2.0
    Microsoft.Bcl.Memory
        System.Memory
            System.Buffers
            System.Numerics.Vectors
            System.Runtime.CompilerServices.Unsafe
        System.Runtime.CompilerServices.Unsafe
```

nguerrera:
net9.0+ would have zero dependencies. 

On .net8.0 Microsoft.Bcl.Memory is used for `Base64Url`, which is in the box in net9.0+.

On netstandard2.0, we are taking advantage of the dependencies of Microsoft.Bcl.Memory for several addtiional things including `Span`.

michaelcfanning:
Looks solid.

michaelcfanning (approved):
:shipit:

Issue #56, opened 2025-03-30
Implementation allows for invalid month + day combinations


michaelcfanning:
e.g., we permit a month value of `B` (February) to be paired with a day value of `d` (29) in a non-leap year situation.

A truly complete solution would account for the Cask-legal leap years between 2028 - 2088.

We do not have a timestamp range that requires handling a 'century' leap year'.

nguerrera:
I'm trying to give up on arguing against "human/regex readable base64 timestamps", but if we decide this is an important concern, then I will find renewed energy to try to push back again. 😅

Since I don't think I actually wrote it down anywhere easy to find, let me write this here...

My preference for the timestamp, which we rejected, was for it to be "an integer number of units from an epoch". Probably more specifically, "number of seconds since Jan 1, 2025." You can compute this easily anywhere unix timestamps are already used, which is basically everywhere. We'd read/write this from the bytewise representation without caring about what it looks like in base64url.

I would like us to write down in the spec why it's important that you can decipher the timestamp with your eyes or a regex. We're paying *a lot* of complexity for it already and if we must restrict to valid month/day combos, we'd pay even more for it.

nguerrera:
Ha, yes, I suppose. I should avoid getting too far into that as long as we haven't decided to go in that direction, and we haven't. 

I'm not an expert on this topic, but I think we can say that leap seconds are not taken into account, which matches https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimeseconds?view=net-9.0

> It does not take leap seconds into account.

And

https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap04.html#tag_21_04_16

This one is more entertaining but basically says to not take leap seconds into account while acknowledging dissenting opinions.

We could also say that it is implementation-defined. That way, you can do whatever your available API to get a unix timestamp does, and we just accept some imprecision.

michaelcfanning:
More thoughts based on my own offline research. I will now go look at Nick's information to see if everything lines up.

- Leap second determinations are made apparently in sufficient time to describe whether they will occur every June or December (and can't be predicated further ahead in time, apparently).
[...]

PR #62, opened 2025-04-09
YAFC: Remove correlating id. Expand and move timestamp.

### LINKED ISSUE #60: YAFC: yet another format change (eliminate public id, move timestamp, enable accuracy for streaming)
In our attempts to design the most secure remediation system possible using the latest CASK design (which includes a non-sensitive correlating id to use as a key to secrets metadata), we have realized that the use of the correlating id causes proof-of-possession problems for us. Specifically, because the sensitive data of a key is entirely decoupled from the non-sensitive id (in the current design), any receiver that performs a security sensitive operation can't require the caller to present a secret in its entirety before trusting that operation should occur. The very obvious case is quarantining an exposed secret on receiving an exposure report.

Based on this observation, we will be removing the non-sensitive correlating id from the format.

Separately (and prompting the creation of #58), we would like to ensure that it is possible to efficiently scan non-delimited data streams for CASK secrets with a low risk of false positives. One easy way to accomplish this is to return to the use of a CRC32 checksum at a fixed location in the key. Another way (that does not require adding a CRC32 check again) is to simply replace some of the space we will recover by dropping the correlating id with padding and/or other data that increases the overall accuracy with a simple pattern match.

Dropping the non-sensitive id creates a new concern, which is how to truncate the secret in an acceptable way for user-facing error messages. It is helpful in reporting scenarios to retain a small segment of the secret in the output (e.g., in an IDE error message) so that the user can actively search for the finding in a document. Previously, the non-sensitive id was perfect for this purpose: on removal we need another solution.

This issue proposes to address all of the above with the following changes:

- Drop the 15 bytes of non-sensitive correlating id at the end of the key.
- Expand the timestamp to include the seconds data and move it to the end of the key (with 12 bits of leading padding to bring the overall timestamp segment to comprise 6 bytes). The timestamp itself can now serve as a non-sensitive identifier in some UX contexts. The leading fixed padding `AA` reduces the overall chance of false positives scanning a randomized, non-delimited stream of bytes encoded as base64 data.
- Expand the fixed `QJJQ` signature to include a reserved character following it. This encoded value (base64 index 0 encoded as `A`) can effectively comprise a version identifier and further reduces the overall chance of false positives in non-delimited streaming scenarios.

The changes above are relatively small to what we have today. The additional fixed padding brings FPs rates for the streaming scenario down to a very acceptable level (see #58 for more). The seconds data allows for some additional key granularity (potential useful for CASK utilization in short-lived secret scenarios). Moving the timestamp to the end of the secret will allow for a decrease in regex size.

michaelcfanning:
- This change removes the correlating id. On an internal security review of this data, we do not believe it is useful to include as a standard part of the format.
- We have expanded the timestamp to include seconds data and moved it to the rightmost portion of the key. This both increases the granularity of the timestamp (useful for short-lived secret scenarios) and also effectively creates a non-sensitive identifier that can be used in error reporting, etc., to describe a finding. In VS, for example, we would truncate the secret and display only the rightmost six characters of the key (i.e., the timestamp) in the error-list. The benefit of retaining a small (non-sensitive) portion of the secret in a user-facing string is that this data can be used to search the original scan target for the exposure, can be used in email/chat communications about an exposure, etc.
- We have sprinkled additional padding in the format that brings detection accuracy to extremely high levels. This allows us to avoid techniques such as incorporating a checksum into key production/validation. A checksum is not inherently undesirable but we are trying to minimize supporting API required to generate and validate CASK secrets.
- I have added exhaustive tests that ensure all padding is required to be present for a key to be deemed valid.

This PR resolves #60. We expressly have chosen *not* to move forward with a fixed header/footer solution (as discussed in #48) at this time. See that issue for deeper discussion on open questions.


src/Cask/CaskKey.cs
```diff
@@ -174,7 +174,7 @@ private void ThrowIfNotInitialized()
     // language=regex
     private const string RegexPattern =
         """
-        (^|[^A-Za-z0-9+\/\-_])([A-Za-z0-9\-_]{24}QJJQB|[A-Za-z0-9\-_]{44}QJJQC|[A-Za-z0-9\-_]{64}QJJQD|[A-Za-z0-9\-_]{88}QJJQE)(A[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{24}|B[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{28}|C[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{32}|D[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{36}|E[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-9\-_][A-Za-z0-9\-_]{40})([^A-Za-z0-9+\/\-_]|$)
+        (^|[^A-Za-z0-9+\/\-_])([A-Za-z0-9\-_]{21}[AQgw]AAQJJQAB|[A-Za-z0-9\-_]{42}[AEIMQUYcgkosw048]AQJJQAC|[A-Za-z0-9\-_]{64}QJJQAD|[A-Za-z0-9\-_]{85}[AQgw]AAQJJQAE)(A[A-Za-z0-9\-_]{5}|B[A-Za-z0-9\-_]{9}|C[A-Za-z0-9\-_]{13}|D[A-Za-z0-9\-_]{17}|E[A-Za-z0-9\-_]{21})AA[A-Za-z0-9\-_][A-L][A-Za-e][A-X][A-Za-z0-7][A-Za-z0-7]([^A-Za-z0-9+\/\-_]|$)
```
michaelcfanning, line 177: >(^|[^ [](http://example.com/codeflow?start=8&length=5)

Haha! It's shorter than it was! This regex is extremely capable. I can improve it further (and irritate others further potentially) by eliminating some illegal date values in the regex itself. :) Seriously, I will provide a more readable regex to use with some other post-processing soon (as a new benchmark!).

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -846,10 +887,10 @@ public void CaskSecrets_GenerateKey_NotDeterministic(SecretSize secretSize)
     }
 
     [Theory]
-    [InlineData(SecretSize.Bits128, "_____________________wAAQJJQBBAAAAAMABCDTEST____________________")]
-    [InlineData(SecretSize.Bits256, "__________________________________________8AQJJQCBAAAAAMABCDTEST____________________")]
-    [InlineData(SecretSize.Bits384, "________________________________________________________________QJJQDBAAAAAMABCDTEST____________________")]
-    [InlineData(SecretSize.Bits512, "_____________________________________________________________________________________wAAQJJQEBAAAAAMABCDTEST____________________")]
+    [InlineData(SecretSize.Bits128, "_____________________wAAQJJQABBMTESTABCDAAAAAAAA")]
```
michaelcfanning, line 890: >_____________________wAAQJJQABBMTESTABCDAAAAAAAA [](http://example.com/codeflow?start=37&length=48)

concretely, we can see that dropping the correlating id as required drops key size significantly. useful.

src/Cask/Cask.cs
```diff
@@ -100,74 +99,100 @@ public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)
...
+        }
 
-        // Any encoded provider key kind, i.e.,
-        // dataLengthsTimeStampProviderKeyKindChars[7], is legal.
+        char second = paddingAndTimestampChars[7];
+        // An encoded second (a zero-indexed value) that exceeds 59 ('7') is not valid.
+        return
+            (second >= 'A' && second <= 'Z') ||
+            (second >= 'a' && second <= 'z') ||
+            (second >= '0' && second <= '7');
```
nguerrera, line 187: IsCask*Bytes* is now dominated by interpreting base64url-encoded chars. This is very confusing to me and possibly also leaving some performance on the table.
  
We have: IsCask(string) -> Decode -> IsCask(bytes) -> Encode subset, check details.

Without the checksum, is there any reason remaining to produce the full set of bytes to validate text or would it make more sense to validate text directly at this point?

There is risk of more code duplication here, but I have some ideas on how to address that and come away with simpler and faster code without any changes to the format. It will be easiest for me to try this and send a PR following this one if it works out. I will do that.

src/Cask/CaskKey.cs
nguerrera: > `AEIMQUYcgkosw048`

Ah, yes of course. 😂 

> and irritate others further potentially

I am irritated by this regex and yes, I will be irritated further if it gets any more complex. 

Assuming a regex of this complexity makes it to v1.0, I would like to investigate generating it from human-readable code that more obviously follows from the format specification.

nguerrera:
Overall, looks good. None of my comments are blocking. I'll do another pass on anything that changes when this is taken out of draft.

docs/CaskSecret.md
```diff
@@ -44,23 +46,19 @@
...
+                                                                            ; facilitating both readability of encoded form and byte-wise use.
+<timestamp> ::= <year> <month> <day> <hour> <minute> <seconds>              ; Time-of-allocation components.
+<year> ::= <base64url>                                                      ; Year of allocation, 'A' (2025) to '_' (2088).
+<month> ::= 'A'..'L'                                                        ; Month of allocation, 'A' (January) to 'L' (December).
+<day> ::= 'A'..'Z' | 'a'..'e'                                               ; 'A' = day 1, 'B' = day 2, ... 'e' = day 31
+<hour> ::= 'A'..'X'                                                         ; Hour of allocation, 0-23. 'A' = hour 0 (midnight), ... 'X' = hour 23.
+<minute> ::= 'A'..'7'                                                       ; Minute of allocation, 0-59.
+<second> ::= 'A'..'7'                                                       ; Second of allocation, 0-59.
```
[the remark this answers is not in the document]
nguerrera: Related: #56

src/Cask/Cask.cs
nguerrera: OK one thing here is we're relying on the Decode step to check if valid base64 with no padding or whitespace and direct validation of text must include that somehow.

This is all not related to this change, and non-blocking.

docs/CaskSecret.md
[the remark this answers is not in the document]
michaelcfanning: OK, having had leap seconds brought to me attention in the last 24 hours, my internet research reveals the following relevant facts:

- Leap second determinations are made apparently in sufficient time to describe whether they will occur every June or December (and can't be predicated further ahead in time, apparently).
- There is agreement to deprecate leap seconds by 2035. Interestingly, is not because we have line of sight to a solution, but because our increased precision in timing at the atomic level is useful (and unignorable) on the one hand but (on the other) presents unacceptable disruptions to systems, e.g., that aren't easily designed to obtain a message across the network that a second should be added or elided every six months. And so we're just going to ignore the problem and let the discrepancy accumulate over 100 years.
- Climate change appears to be impacting rate of orbiting, compounding the issue. I read that this impact may trend towards negative leap seconds as opposed to adding one).

Based on the decision of various standards bodies to literally drop the complexity of leap seconds, I suggest that we conveniently ignore the matter as well. 

Again, casting about on a topic that's new time, it appears for Windows at least that we'd never actually expect to see a time with a wonky seconds value. Instead, what happens is that the Windows time service sets a leap indicator that indicates whether a second should be lost or added (i.e., it does not raise an 'extra' notification with an odd seconds value). The UTC time subsequently refreshes the current time in the next notification, if necessary. .NET itself picks up its notions of time from the Windows time service so itself won't produce an encoded value with a 61st seconds value.

[https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/time-service-treats-leap-second](https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/time-service-treats-leap-second)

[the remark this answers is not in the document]
michaelcfanning: Closing this comment. We can use the issue if necessary to continue the leap seconds conversation.

docs/CaskSecret.md
```diff
@@ -2,14 +2,16 @@
...
+          <reserved>            ; 6 bits of reserved padding.
           <sensitive-data-size> ; A count of 16-byte segments encoded as sensitive data ('B' = 1 x 16 bytes = 128 bits, etc).
           <optional-data-size>  ; A count of 3-byte optional data segments, 'A' = 0 = 0 bytes, 'B' = 1 = 3 bytes, etc.
-          <timestamp>           ; The year, month, day, hour, and minute of secret allocation.
           <provider-kind>       ; A provider-defined key kind.
-          [<optional-fields>]   ; Optional fields comprising provider-defined data.
           <provider-signature>  ; A fixed signature identifying the secret provider.
-          <correlating-id>      ; A 20-byte non-sensitive correlating identifier for the secret (see below).
+          [<optional-fields>]   ; Optional 3-byte segments comprising provider-defined data.
+          <reserved>            ; 12 bits of reserved padding.
```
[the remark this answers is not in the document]
michaelcfanning: yes! good note.

src/Cask/Cask.cs
michaelcfanning: OK, I strongly support giving this topic some attention.

Could we simply prefer to invert the process? i.e., in `IsBytes`, encode as a string and pass to helpers that validate the encoded form. If we took this approach, we will need to keep your comment in mind from the other direction: the textual validator would itself need to ensure it was passed valid base64 url printable chars only, of correct length.

src/Cask/CaskKey.cs
nguerrera: I look forward to that, but my comment here was about the maintainability, not the performance. If perf is good, then I still want to see something done to make this more readable.

Another subtlety: IsCask has picked up signifcantly more complexity as the regex has simultaneously gotten more complicated. We should sanity check the perf of IsCask is still good. Intuively, I don't expect anything that's been added to be overly expensive, but we should make sure of that when comparing simple regex + IsCask to complex regex.

Even more subtle: some choices like fixed footer that make solving the whole thing with complex regex difficult or impossible might also make IsCask simpler and faster

Anyway, resolving this, it's all separate.

src/Cask/Cask.cs
nguerrera: I filed #64

src/Cask/Cask.cs
```diff
@@ -210,77 +250,81 @@ public static CaskKey GenerateKey(string providerSignature,
...
+        destination = destination[PaddingAndTimestampSizeInBytes..];
 
         return CaskKey.Encode(key);
     }
 
-
-    private static Range ComputeCaskSignatureCharRange(SecretSize secretSize)
+    private static Range ComputeCaskSignatureCharRange(int keyLengthInChars, out SecretSize secretSize)
     {
+        int keyLengthInBytes = keyLengthInChars / 4 * 3;
```
nguerrera, line 294: We have a helper for this, `Base64CharsToBytes` but I guess here the reason it is not used is because it handles unaligned input sizes and here we can assert that the keyLengthInChars is already validated?

Possibly some opportunities to also clean things like this up with #64.

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -257,45 +304,32 @@ private void TestEncodedMatchedDecoded(string encodedKey,
...
+        char encodedHourChar = ppymdhmsChars[5];
+        char encodedMinuteChar = ppymdhmsChars[6];
+        char encodedSecondChar = ppymdhmsChars[7];
 
         int year = base64UrlPrintableCharIndices[encodedYearChar];
         int month = base64UrlPrintableCharIndices[encodedMonthChar];
         int day = base64UrlPrintableCharIndices[encodedDayChar];
         int hour = base64UrlPrintableCharIndices[encodedHourChar];
         int minute = base64UrlPrintableCharIndices[encodedMinuteChar];
+        int second = base64UrlPrintableCharIndices[encodedMinuteChar];
```
nguerrera, line 323:
```suggestion
        int second = base64UrlPrintableCharIndices[encodedSecondChar];
```

This is the first objective issue I've found on this PR.😊

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -652,6 +691,44 @@ public void CaskSecrets_IsCask_AllMinutes(SecretSize secretSize)
...
+            Span<char> destination = key.ToCharArray().AsSpan();
+            char encodedSecond = Base64UrlChars[base64Index];
+            destination[minuteCharOffset] = encodedSecond;
+
+            string modifiedKey = destination.ToString();
+            valid = Cask.IsCask(modifiedKey);
+
+            if (expectedValid)
+            {
+                Assert.True(valid, $"'IsCask' unexpectedly failed after modifying Cask timestamp with valid minute '{encodedSecond}': {modifiedKey}");
```
nguerrera, line 722:
```suggestion
                Assert.True(valid, $"'IsCask' unexpectedly failed after modifying Cask timestamp with valid second '{encodedSecond}': {modifiedKey}");
```

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -652,6 +691,44 @@ public void CaskSecrets_IsCask_AllMinutes(SecretSize secretSize)
...
+                IsCaskVerifySuccess(modifiedKey);
+                continue;
+            }
+
+            Assert.False(valid, $"'IsCask' unexpectedly succeeded after modifying Cask timestamp with invalid minute '{encodedSecond}': {modifiedKey}");
```
nguerrera, line 727:
```suggestion
            Assert.False(valid, $"'IsCask' unexpectedly succeeded after modifying Cask timestamp with invalid second '{encodedSecond}': {modifiedKey}");
```

nguerrera (approved):

src/Cask/Cask.cs
michaelcfanning: I think probably converging on the helper might make sense.

Issue #64, opened 2025-04-11
Avoid encode/decode round-trips in IsCask


nguerrera:
See https://github.com/microsoft/cask/pull/62#discussion_r2037629845

PR #66, opened 2025-04-16
Make IsCask and IsCaskUtf8 implementations consistent

nguerrera:
`IsCask` taking UTF-16 validated the secret size up-front while `IsCaskUtf8` deferred to `IsCaskBytes` to do so. I noticed this discrepancy while looking at #64 where I am experimenting with tricks for sharing code between UTF-8 and UTF-16 cases.

The fix here is to let `IsCaskBytes` handle it in both cases, which is consistent with several other fields that are only validated after decoding to bytes.

This change also moves `IsCaskUtf8` next to `IsCask` to emphasize that these are the same operation for different encodings, to make it easier to look for unwanted differences between them, and to keep public API above helpers.

Finally, parameter names and method documentation are updated to help communication the distinction between the binary decoded form and the base64url-encoded textual forms.

### DIFF
--- a/src/Cask/Cask.cs
+++ b/src/Cask/Cask.cs
@@ -69,6 +69,48 @@ public static class Cask
         return IsCaskBytes(keyBytes);
     }
 
+
+    /// <summary>
+    /// Validates that the provided UTF8-encoded byte sequence represents a valid Cask key.
+    /// </summary>
+    public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUtf8)
+    {
+        if (!IsValidKeyLengthInChars(keyUtf8.Length))
+        {
+            return false;
+        }
+
+        Range caskSignatureCharRange = ComputeCaskSignatureCharRange(keyUtf8.Length, out SecretSize _);
+
+        // Check for CASK signature, "QJJQ".
+        if (!keyUtf8[caskSignatureCharRange].SequenceEqual(CaskSignatureUtf8))
+        {
+            return false;
+        }
+
+        int lengthInBytes = Base64CharsToBytes(keyUtf8.Length);
+        Debug.Assert(lengthInBytes <= MaxKeyLengthInBytes);
+        Span<byte> keyBytes = stackalloc byte[lengthInBytes];
+
+        OperationStatus status = Base64Url.DecodeFromUtf8(keyUtf8,
+                                                          keyBytes,
+                                                          out int charsConsumed,
+                                                          out int bytesWritten,
+                                                          isFinalBlock: true);
+
+        Debug.Assert(status is OperationStatus.InvalidData || charsConsumed == keyUtf8.Length);
+        Debug.Assert(status is not OperationStatus.DestinationTooSmall or OperationStatus.NeedMoreData);
+
+        // NOTE: Decoding can succeed with `bytesWritten < lengthInBytes` if the
+        //       input has padding or whitespace, which we don't allow.
+        if (status != OperationStatus.Done || bytesWritten != lengthInBytes)
+        {
+            return false;
+        }
+
+        return IsCaskBytes(keyBytes);
+    }
+
     /// <summary>
     /// Validates that the provided byte sequence represents a valid Cask key in binary decoded form.
     /// </summary>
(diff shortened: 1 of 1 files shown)


src/Cask/Cask.cs
```diff
@@ -23,22 +23,16 @@ public static bool IsCask(string key)
     }
 
     /// <summary>
-    /// Validates that the provided UTF16-encoded text sequence represents a
-    /// valid Cask key.
+    /// Validates that the provided UTF16-encoded text represents a valid Cask key.
```
nguerrera, line 26: Also made the comments consistents, dropping sequence and saying `UTF16-encoded text` or `UTF8-encoded text`. I think we want to emphasize that this is text (to distinguish it from IsCaskBytes so `UTF8-encoded *byte* sequence` was confusing me.

src/Cask/Cask.cs
```diff
@@ -23,22 +23,16 @@ public static bool IsCask(string key)
...
     /// </summary>
-    /// <param name="key"></param>
     public static bool IsCask(ReadOnlySpan<char> key)
```
michaelcfanning, line 28: >key [](http://example.com/codeflow?start=49&length=3)

maybe we could use `encodedKey` as a hint?

src/Cask/Cask.cs
```diff
@@ -69,6 +63,47 @@ public static bool IsCask(ReadOnlySpan<char> key)
         return IsCaskBytes(keyBytes);
     }
 
+    /// <summary>
+    /// Validates that the provided UTF8-encoded text represents a valid Cask key.
+    /// </summary>
+    public static bool IsCaskUtf8(ReadOnlySpan<byte> keyUtf8)
```
michaelcfanning, line 69: >keyUtf8 [](http://example.com/codeflow?start=53&length=7)

again, maybe 'encodedKey' is a help? just a thought. naming is hard.

michaelcfanning:
>     public static bool IsCaskBytes(ReadOnlySpan<byte> keyBytes)

`decodedKey`?

---
Refers to: src/Cask/Cask.cs:110 in 7ff8f5d. [](commit_id = 7ff8f5d0bbebfb9d6011ac7a20a4eaf906012b53, deletion_comment = False)

src/Cask/Cask.cs
```diff
@@ -69,6 +63,47 @@ public static bool IsCask(ReadOnlySpan<char> key)
...
+        if (status != OperationStatus.Done || bytesWritten != lengthInBytes)
+        {
+            return false;
+        }
+
+        return IsCaskBytes(keyBytes);
+    }
+
     /// <summary>
     /// Validates that the provided byte sequence represents a valid Cask key in binary decoded form.
```
michaelcfanning, line 108: >binary decoded form [](http://example.com/codeflow?start=81&length=19)

just as you note here that that this method is for the decoded form, maybe add a similar comment to the others (that the data is base64url-encoded)

src/Cask/Cask.cs
nguerrera: Done.

src/Cask/Cask.cs
nguerrera: Good idea. Done.

src/Cask/Cask.cs
```diff
@@ -23,40 +23,77 @@ public static bool IsCask(string key)
...
+        Debug.Assert(lengthInBytes <= MaxKeyLengthInBytes);
+        Span<byte> keyBytes = stackalloc byte[lengthInBytes];
+
+        OperationStatus status = Base64Url.DecodeFromUtf8(encodedKey,
+                                                          keyBytes,
+                                                          out int charsConsumed,
+                                                          out int bytesWritten,
+                                                          isFinalBlock: true);
+
+        Debug.Assert(status is OperationStatus.InvalidData || charsConsumed == encodedKey.Length);
```
michaelcfanning, line 96: >status [](http://example.com/codeflow?start=21&length=6)

You are gtg but can you check the code coverage to see whether we are still at 100%? tx

michaelcfanning (approved):
:shipit:

michaelcfanning (approved):

src/Cask/Cask.cs
nguerrera: There's a pre-existing less than 100% coverage issue. I will fix it separately since it's unrelated.

nguerrera: #67

PR #69, opened 2025-04-17
Remove C++ implementation stub

nguerrera:
We will add this back when we're actually ready to work on the C++ implementation.


michaelcfanning (approved):
:shipit:

PR #71, opened 2025-04-17
Limit key sizes to 256 or 512 bits.

### LINKED ISSUE #63: Limiting key size to maximize optional data size
Currently, due to the absence of a fixed header/footer (as proposed in #48), we limit the size of optional data in a CASK key so that the size of the sensitive data itself can be inferred from the encoded key length. 

The limit is currently a maximum of 12 bytes of optional data (strictly speaking, due to the nature of base64-encoding, 15 bytes are possible for a 384-bit key before its length equals the minimal length of 512-bit key. We stick with a 12 bytes limit simply to be consistent).

12 bytes on its own has at least one obvious limit: if an allocator cared to generate and encode an additional unique id in the secret, 16 bytes are generally recommended for reliable uniqueness. Since this is CASK, a more pragmatic choice might be 15 bytes, as this is a 3-byte aligned value (the next step up would require 18). 120 bits of entropy provides an acceptable level of uniqueness (it's worth noting that a GUID, which is designed for uniqueness, has 6 reserved bits and therefore can only allocate 122 bits to the uniqueness property).

If we 'simplified' CASK to only permit either 256- or 512- bit secrets, we would increase the maximum permissible optional data size to 33 bytes for a 256-bit key (because a 256-bit key with maximal optional data would still have a key length less than the minimal 512 bit key).

Minimal 512-bit key: 66 bytes entropy + 15 bytes required CASK data == 81 bytes.
Maximal 256-bit key: 33 bytes entropy + 15 bytes required + 30 bytes option == 78 bits

This change would also allow us to eliminate the optional data size constraint entirely for a 512-byte key (because it wouldn't be 'competing' with a key of larger size. Balancing this, we might compromise flexibility in versioning the format, i.e., to accommodate a larger key size, we'd need to version the format to enforce an optional data size limit. This could be a serious concern if users had already shipped CASK keys exceeding the vNext retrofitted size constraint. And so this probably is not a good option.

Realistically, 30 bytes of optional data is a good limit, basically permitting up to two entirely unique ids, or a single unique id plus up to 15 bytes of additional arbitrary data. 

A downside here is key size across the wire. Although 128-bit keys are currently considered unbreakable in most contexts, a larger key size is recommended by many to future-proof for a post-quantum world, as additional security for extremely sensitive environments, operations, etc. And so requiring a CASK key producer that chooses to only generate 128 bits of entropy (but still package it in the 256-bit form factor) could be acceptable. Similarly, a producer with a 384-bit scenario (as some cryptographic hashing algorithms produce) would be required to transport that data in a 512-bit CASK key form factor.

michaelcfanning:
Resolves #63.

We limit valid key size to 32 or 64 bytes (padded to 33 and 66 bytes respectively) of sensitive data. We now buy the ability to increase optional size component to 30 bytes (while retaining the ability to infer sensitive data and optional data size directly from key length).

After accepting this format change, the format is considered approved by the core design team and ready for expanded review. We will author a reviewable specification next that captures current state.


src/Tests/Cask.Tests/CaskKeyTests.cs
```diff
@@ -47,7 +43,7 @@ public void CaskKey_SizeInBytes(string? providerData)
 
         foreach (SecretSize secretSize in CaskTestsBase.AllSecretSizes)
         {
-            int secretSizeInBytes = (int)secretSize * 16;
+            int secretSizeInBytes = (int)secretSize * 32;
```
michaelcfanning, line 46: >32 [](http://example.com/codeflow?start=54&length=2)

it might be useful to define a shared test constants library (or to simply pick up this information directly from the product).

src/Cask/Cask.cs
```diff
@@ -429,24 +414,12 @@ private static void ValidateTimestamp(DateTimeOffset now)
...
     private static bool IsValidKeyLengthInChars(int length)
     {
-        if (length < MinKeyLengthInChars || length > MaxKeyLengthInChars || !Is4CharAligned(length))
-        {
-            return false;
-        }
-
-        // There is gap in valid key lengths between 384-bit and 512-bit keys.
-        return length <= Max384BitKeyLengthInChars || length >= Min512BitKeyLengthInChars;
+        return !(length < MinKeyLengthInChars || length > MaxKeyLengthInChars || !Is4CharAligned(length));
```
nguerrera, line 417: I think it's a little easier to understand this way:

```suggestion
        return length >= MinKeyLengthInChars && length <= MaxKeyLengthInChars && Is4CharAligned(length);
```

src/Cask/Cask.cs
```diff
@@ -429,24 +414,12 @@ private static void ValidateTimestamp(DateTimeOffset now)
...
     private static bool IsValidKeyLengthInBytes(int length)
     {
-        if (length < MinKeyLengthInBytes || length > MaxKeyLengthInBytes || !Is3ByteAligned(length))
-        {
-            return false;
-        }
-
-        // There is a gap in valid key lengths between 384-bit and 512-bit keys.
-        return length <= Max384BitKeyLengthInBytes || length >= Min512BitKeyLengthInBytes;
+        return !(length < MinKeyLengthInBytes || length > MaxKeyLengthInBytes || !Is3ByteAligned(length));
```
nguerrera, line 422:
```suggestion
        return length >= MinKeyLengthInBytes && length <= MaxKeyLengthInBytes && Is3ByteAligned(length));
```

src/Tests/Cask.Tests/CaskSecretsTests.cs
```diff
@@ -79,10 +68,10 @@ private void TestEncodedMatchedDecoded(string encodedKey,
         ReadOnlySpan<char> encodedSource = encodedKey.AsSpan();
         ReadOnlySpan<byte> decodedSource = decodedKey;
 
-        // A CASK secret may encode 128, 256, 384, or 512 bits of sensitive data
-        // and its length will differ accordingly. CASK also allows for optional
+        // A CASK secret may encode 256 or 512 bits of sensitive data and and
```
nguerrera, line 71:
```suggestion
        // A CASK secret may encode 256 or 512 bits of sensitive data and
```

nguerrera (approved):

src/Cask/Cask.cs
michaelcfanning: thank you for this transformation in propositional logic

nguerrera: Don't thank me, thank De Morgan.

src/Cask/Cask.cs
nguerrera: doh, sorry my suggestion didn't get the parens right.

nguerrera (approved):

